PR #5, opened 2024-06-14
Script to bootstrap CI AWS pipeline

khluu:
- Always trigger tests if it's on `main` branch
- Check if ACM files matched with one of the patterns to trigger tests.
- Otherwise, skip uploading pipeline and exit 0.


khluu:
Tested here: https://buildkite.com/vllm/ci-aws/builds/1430#01901567-aceb-41b1-909f-ae5740df150b the first `diff check` step starts the pipeline as expected

PR #8, opened 2024-06-25
Share HuggingFace downloads between test runs

DarkLight1337:
The model tests take over 50 minutes, which is quite long and runs the risk of getting interrupted. This PR (ported from vllm-project/vllm#4874) attempts to reduce the running time by sharing the HuggingFace cache between test runs so that models need not be downloaded each time. 

Please share any concerns you may have regarding this approach. I'm also not sure how to test the resulting speed-up since there is no guarantee that model tests are re-run in the same machine (and hence able to utilize the cache effectively).

**Note:** `hostPath` volumes in Kubernetes have associated security risks. Is there another way for [agent-stack-k8s](https://github.com/buildkite/agent-stack-k8s/tree/main) to use a persistent volume?


DarkLight1337:
I haven't noticed any significant improvement for `models-tests` so far...

PR #10, opened 2024-07-01
Fix apt install bug

KuntaiDu:
The current CI triggers an error:
image
So I remove the line `apt install xxxxx`.


KuntaiDu:
This bug has been fixed by Kevin. Close this PR.

PR #15, opened 2024-07-11
[CI/Build] (1/2) Switching AMD CI to store images in Docker Hub

adityagoel14:
1st of 2 PRs that is required to switch AMD CI to using Docker Hub. This adds a build step where the machine logs into docker, builds the image, then pushes it to a public Docker Hub registry. 

On the Buildkite side, a new queue called "amd-cpu" needs to be created. 

This PR must be merged first and does not require the second PR to be merged. This also serves as a way to confirm that the push is working correctly.


scripts/test-template-aws.j2
```diff
@@ -177,4 +199,4 @@ steps:
               path: {{ hf_home }}
               type: Directory
   {% endif %}
-  {% endfor %}
+  {% endfor %}
```
simon-mo, line 202:
```suggestion
  {% endfor %}

```

PR #20, opened 2024-07-15
deal with cases where nvidia-smi is not available

youkaichao:


youkaichao:
found in https://buildkite.com/vllm/ci-aws/builds/4904#0190b3ae-20ce-4ea4-ac4d-23ec096b6c9d

PR #28, opened 2024-08-30
Add ppc64le cpu test job

sumitd2:
To be merged after https://github.com/vllm-project/vllm/pull/8026


simon-mo:
@khluu^ I merged the vLLM PR but will leave this up to you.

khluu:
@sumitd2 I'm going to merge it but mark it with soft-fail first so it doesn't block CI just in case, since this is a new test ... I'll revert the soft fail once the test is working consistently!

PR #30, opened 2024-09-17
New Python-based Buildkite pipeline generator

khluu:
As our [jinja template](https://github.com/vllm-project/buildkite-ci/blob/main/scripts/test-template-aws.j2) is getting messy/limited, this Buildkite pipeline generator implemented w/ Python hopefully can extend our flexibility with how we setup our CI pipeline.

This PR implements the pipeline generator and its components. The test changes are in https://github.com/vllm-project/vllm/pull/8324. The 2 PRs should be merged at the same time

### DIFF
--- a/scripts/ci_aws_bootstrap.sh
+++ b/scripts/ci_aws_bootstrap.sh
@@ -5,7 +5,17 @@ set -euo pipefail
 if [[ -z "${RUN_ALL:-}" ]]; then
     RUN_ALL=0
 fi
-
+export VLLM_BUILDKITE_BRANCH="pipeline_gen"
+generate_pipeline() {
+    python -m pip install click pydantic
+    curl -o .buildkite/pipeline_generator.py https://raw.githubusercontent.com/vllm-project/buildkite-ci/$VLLM_BUILDKITE_BRANCH/scripts/pipeline_generator/pipeline_generator.py
+    curl -o .buildkite/plugin.py https://raw.githubusercontent.com/vllm-project/buildkite-ci/$VLLM_BUILDKITE_BRANCH/scripts/pipeline_generator/plugin.py
+    curl -o .buildkite/step.py https://raw.githubusercontent.com/vllm-project/buildkite-ci/$VLLM_BUILDKITE_BRANCH/scripts/pipeline_generator/step.py
+    curl -o .buildkite/utils.py https://raw.githubusercontent.com/vllm-project/buildkite-ci/$VLLM_BUILDKITE_BRANCH/scripts/pipeline_generator/utils.py
+    python .buildkite/pipeline_generator.py --run_all=$RUN_ALL --list_file_diff="$LIST_FILE_DIFF"
+    buildkite-agent pipeline upload .buildkite/pipeline.yaml
+    exit 0
+}
 upload_pipeline() {
     echo "Uploading pipeline..."
     ls .buildkite || buildkite-agent annotate --style error 'Please merge upstream main branch for buildkite CI'
@@ -74,4 +84,4 @@ LIST_FILE_DIFF=$(get_diff | tr ' ' '|')
 if [[ $BUILDKITE_BRANCH == "main" ]]; then
     LIST_FILE_DIFF=$(get_diff_main | tr ' ' '|')
 fi
-upload_pipeline
+generate_pipeline
--- a/scripts/ci_aws_bootstrap.sh
+++ b/scripts/ci_aws_bootstrap.sh
@@ -13,6 +13,7 @@ generate_pipeline() {
     curl -o .buildkite/step.py https://raw.githubusercontent.com/vllm-project/buildkite-ci/$VLLM_BUILDKITE_BRANCH/scripts/pipeline_generator/step.py
     curl -o .buildkite/utils.py https://raw.githubusercontent.com/vllm-project/buildkite-ci/$VLLM_BUILDKITE_BRANCH/scripts/pipeline_generator/utils.py
     python .buildkite/pipeline_generator.py --run_all=$RUN_ALL --list_file_diff="$LIST_FILE_DIFF"
+    cat .buildkite/pipeline.yaml
     buildkite-agent pipeline upload .buildkite/pipeline.yaml
     exit 0
 }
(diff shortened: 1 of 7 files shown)


scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
 
 set -euo pipefail
 
-if [[ -z "${RUN_ALL:-}" ]]; then
-    RUN_ALL=0
-fi
+RUN_ALL=${RUN_ALL:-0}
+export VLLM_BUILDKITE_BRANCH="pipeline_gen"
```
khluu, line 6: the branch here would be changed to `main` after merge

aslonnie: you can set this to `VLLM_BUILDKITE_BRANCH=${VLLM_BUILDKITE_BRANCH:-main}` and then you can merge it and manually set `VLLM_BUILDKITE_BRANCH=pipeline_gen` or `your-own-branch` when doing testing.

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
+
+generate_pipeline() {
+    python -m pip install click pydantic
```
aslonnie, line 9: pin the versions of the packages?

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
+    
+    # Download necessary files
+    for file in pipeline_generator.py plugin.py step.py utils.py; do
```
aslonnie, line 12: variable like `file` should be upper cases, like `FILE`.

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
+        curl -o .buildkite/$file https://raw.githubusercontent.com/vllm-project/buildkite-ci/$VLLM_BUILDKITE_BRANCH/scripts/pipeline_generator/$file
```
aslonnie, line 13: double quote `$FILE`, like `".buildkite/$FILE"`

aslonnie: also please double quote the URL in general. URL often can contain special characters that conflicts with bash (example: `&`)

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
+    done
+    
+    # Generate and upload pipeline
+    python .buildkite/pipeline_generator.py --run_all=$RUN_ALL --list_file_diff="$LIST_FILE_DIFF"
+    cat .buildkite/pipeline.yaml
+    buildkite-agent pipeline upload .buildkite/pipeline.yaml
+    exit 0
```
aslonnie, line 20: why does it need to `exit 0` here?

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
+}
 
 upload_pipeline() {
     echo "Uploading pipeline..."
     ls .buildkite || buildkite-agent annotate --style error 'Please merge upstream main branch for buildkite CI'
+    
+    # Install minijinja-cli
     curl -sSfL https://github.com/mitsuhiko/minijinja/releases/latest/download/minijinja-cli-installer.sh | sh
     source /var/lib/buildkite-agent/.cargo/env
+
```
aslonnie, line 30: since we are here, use double bracket `[[ ]]` for conditions?

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
+    cat .buildkite/pipeline.yaml
+    buildkite-agent pipeline upload .buildkite/pipeline.yaml
+    exit 0
+}
 
 upload_pipeline() {
     echo "Uploading pipeline..."
     ls .buildkite || buildkite-agent annotate --style error 'Please merge upstream main branch for buildkite CI'
+    
+    # Install minijinja-cli
```
aslonnie, line 27: pin a version, don't use `latest`?

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
+        handle_regular_pipeline
     fi
-    if [ ! -e ".buildkite/test-template.j2" ]; then
+}
+
+handle_fastcheck() {
+    [ ! -e ".buildkite/test-template-fastcheck.j2" ] && \
+        curl -o .buildkite/test-template-fastcheck.j2 https://raw.githubusercontent.com/vllm-project/buildkite-ci/main/scripts/test-template-fastcheck.j2
+    
+    cd .buildkite && minijinja-cli test-template-fastcheck.j2 test-pipeline.yaml > pipeline.yml
```
aslonnie, line 42: this will change the working directory, is that intended?

use `tee` rather than a separate cat? it will allow you to also see the output when the previous line has error.

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
-    echo "Run all: $RUN_ALL"
-    minijinja-cli test-template.j2 test-pipeline.yaml -D list_file_diff="$LIST_FILE_DIFF" -D run_all="$RUN_ALL" > pipeline.yml
-    buildkite-agent pipeline upload pipeline.yml
     exit 0
 }
 
 get_diff() {
-    $(git add .)
-    echo $(git diff --name-only --diff-filter=ACMDR $(git merge-base origin/main HEAD))
+    git add .
```
aslonnie, line 67: why do we need this `git add .`? this will be used on CI, right?

scripts/ci_aws_bootstrap.sh
```diff
@@ -2,76 +2,90 @@
...
+    git diff --name-only --diff-filter=ACMDR $(git merge-base origin/main HEAD)
 }
 
 get_diff_main() {
-    $(git add .)
-    echo $(git diff --name-only --diff-filter=ACMDR HEAD~1)
+    git add .
+    git diff --name-only --diff-filter=ACMDR HEAD~1
```
aslonnie, line 73: this is `git show --pretty="" --name-only` ?

and why `ACMDR` is required? do you know?

aslonnie (commented):
can we split the python files into smaller PR's with unit tests?

aslonnie (commented):
.

PR #31, opened 2024-09-19
Pipeline generator utils

khluu:


scripts/pipeline_generator/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        return AgentQueue.AWS_SMALL_CPU
+    if gpu_type == A100_GPU:
+        return AgentQueue.A100
+    return AgentQueue.AWS_1xL4 if num_gpus == 1 else AgentQueue.AWS_4xL4
+
+
+def get_full_test_command(test_commands: List[str], step_working_dir: str) -> str:
+    """Convert test commands into one-line command with the right directory."""
+    working_dir = step_working_dir or DEFAULT_WORKING_DIR
+    test_commands_str = "; ".join(test_commands)
```
aslonnie, line 42: I am still a bit worried that this is not the write way. maybe concating with end-lines will be safer.

scripts/pipeline_generator/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    return f"cd {working_dir}; {test_commands_str}"
```
aslonnie, line 43: why do you need to add `cd working_dir` here? the docker plugin has builtin `workdir` field?

https://github.com/buildkite-plugins/docker-buildkite-plugin?tab=readme-ov-file#workdiroptional-string

scripts/pipeline_generator/utils.py
```diff
@@ -0,0 +1,62 @@
...
+TEST_PATH = ".buildkite/test-pipeline.yaml"
+EXTERNAL_HARDWARE_TEST_PATH = ".buildkite/external-tests.yaml"
+PIPELINE_FILE_PATH = ".buildkite/pipeline.yaml"
+MULTI_NODE_TEST_SCRIPT = ".buildkite/run-multi-node-test.sh"
+
+STEPS_TO_BLOCK = []
+
+
+class AgentQueue(str, enum.Enum):
+    AWS_CPU = "cpu_queue"
```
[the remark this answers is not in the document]
khluu: the ones with `_queue` are those that we name it ourselves.. the one without are from AMD ..

scripts/pipeline_generator/utils.py
khluu: Added endline char!

scripts/pipeline_generator/utils.py
khluu: I'm not sure if kubernetes plugin has it ... this will be used to generate command for both k8s & docker plugin

khluu: it might work with `workingDir` in the k8s podspec.. I can try it out after the pipeline generator is up and if it works, remove `cd ..` from the command and specify working dir in the plugin

scripts/pipeline_generator/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    if gpu_type == A100_GPU:
+        return AgentQueue.A100
+    return AgentQueue.AWS_1xL4 if num_gpus == 1 else AgentQueue.AWS_4xL4
+
+
+def get_full_test_command(test_commands: List[str], step_working_dir: str) -> str:
+    """Convert test commands into one-line command with the right directory."""
+    working_dir = step_working_dir or DEFAULT_WORKING_DIR
+    test_commands_str = ";\n".join(test_commands)
+    return f"cd {working_dir};\n{test_commands_str}"
```
aslonnie, line 43: if you added the `\n`, do you still need to add the `;` ?

aslonnie (approved):

scripts/pipeline_generator/utils.py
khluu: probably not but let's just leave it there for now... I can test again & take it out later when it's up and running

PR #32, opened 2024-09-19
Pipeline generator steps

khluu:
Define step types and helper functions to help create step.


scripts/pipeline_generator/step.py
```diff
@@ -0,0 +1,60 @@
+from pydantic import BaseModel, Field
+from typing import List, Dict, Any, Optional
+
+from .utils import AgentQueue
+
+
+class TestStep(BaseModel):
+    """This class represents a test step defined in the test configuration file."""
```
aslonnie, line 8: why are these in this PR? do you want to test maybe unmarshaling these data structures?

khluu: the method that unmarshal them from the yaml file would be in the next PR

aslonnie: then why is this checked in in this PR? this is not referenced in this PR anywhere?

aslonnie:
there is a conflict on the `requirements.txt` file

aslonnie (commented):
resolve the conflict? and maybe remove from this PR the data structures that are not referenced?

khluu:
done! I've removed the 2 classes that are not referenced in this PR yet, and also resolved conflict

scripts/pipeline_generator/step.py
```diff
@@ -0,0 +1,28 @@
+from pydantic import BaseModel, Field
+from typing import List, Dict, Any, Optional
+
+from .utils import AgentQueue
+
+
+class BuildkiteBlockStep(BaseModel):
+    """This class represents a block step in Buildkite format."""
+    block: str
+    depends_on: Optional[str] = "build"
```
aslonnie, line 10: maybe declare a global constant for `"build"`?

scripts/pipeline_generator/step.py
```diff
@@ -0,0 +1,28 @@
...
+    key: str
+
+
+def get_step_key(step_label: str) -> str:
+    step_label = step_label.replace(", ", ",")
+    step_key = ""
+    skip_chars = "()%"
+    for char in step_label.lower():
+        if char in ", ":
+            step_key += "-"
```
aslonnie, line 20: maybe this should be like "if the last char was `-`, just skip this char"? then you do not need the `replace(", ", ",")`, and can more gracefully handle cases like `,,` or `,  ` (double spaces)?

aslonnie (approved):

scripts/pipeline_generator/step.py
khluu: done

scripts/pipeline_generator/step.py
khluu: done

PR #33, opened 2024-09-19
Pipeline generator plugins

khluu:
Define Buildkite plugin class for Docker & Kubernetes, along with helper functions to generate these plugin configs.


aslonnie:
could you rebase this on top of master?

scripts/pipeline_generator/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+        "VLLM_USAGE_SOURCE=ci-test",
+        "HF_TOKEN",
+        "BUILDKITE_ANALYTICS_TOKEN"
+    ]
+    volumes: List[str] = [
+        "/dev/shm:/dev/shm",
+        f"{HF_HOME}:{HF_HOME}"
+    ]
+
+class KubernetesPodContainerConfig(BaseModel):
```
aslonnie, line 27: we cancelled the plan to move back to k8s stack? why are we adding this?

khluu: This is for the A100 node on EKS (roblox aws)

aslonnie (commented):
maybe it was lost last time; could you rebase this on top of master?

scripts/pipeline_generator/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+        default=[
+            {"name": "devshm", "emptyDir": {"medium": "Memory"}},
+            {"name": "hf-cache", "hostPath": {"path": HF_HOME, "type": "Directory"}}
+        ]
+    )
+
+class KubernetesPluginConfig(BaseModel):
+    pod_spec: KubernetesPodSpec = Field(alias="podSpec")
+
+def get_kubernetes_plugin_config(docker_image_path: str, test_bash_command: List[str], num_gpus: int) -> Dict:
```
aslonnie, line 71: nit: `docker_image_path` -> `image` or `container_image` ?

k8s does not always use docker. and the image does not have to be docker image.

scripts/tests/pipeline_generator/test_plugin.py
```diff
@@ -0,0 +1,116 @@
...
+
+    assert get_kubernetes_plugin_config(docker_image_path, test_bash_command, num_gpus) == expected_config
+
+
+@pytest.mark.parametrize(
+    "docker_image_path, test_bash_command, no_gpu, expected_config",
+    [
+        (
+            "test_image:latest",
+            ["bash", "-c", "echo A", "pytest -v -s a.py"],
```
aslonnie, line 62: this command looks weird and not real.. "echo A" is the arg of `-c`, waht is the `pytest` after it? why does the command for a docker plugin look like this?

aslonnie (approved):

PR #36, opened 2024-09-25
Pipeline generator

khluu:


khluu:
test build here: https://buildkite.com/vllm/ci-aws/builds/9160
Some tests are failing on main due to recent transformer release (and vllm has to upgrade to support llama3.2 to tests broke)

aslonnie (commented):
talked offline. let's try split this into smaller PR's with clearer modularization / layers.

PR #37, opened 2024-09-28
[1/n][pipeline-gen] Add default test commands

khluu:
- Add default commands that run on every test
- Return 1xL4 queue if there's no `num_gpus` specified


scripts/pipeline_generator/utils.py
```diff
@@ -15,6 +15,12 @@
 PIPELINE_FILE_PATH = ".buildkite/pipeline.yaml"
 MULTI_NODE_TEST_SCRIPT = ".buildkite/run-multi-node-test.sh"
 
+TEST_DEFAULT_COMMANDS = [
+    "(command nvidia-smi || true)",
```
aslonnie, line 19: could you add a comment that explains what does this `nvidia-smi || true` need to be here? it will always pass, so why check?

aslonnie (approved):

PR #39, opened 2024-09-28
[3/n][pipeline-gen] Add TestStep and BuildkiteStep along with validators

khluu:
- Add these 2 step classes to be used in future PRs
- Add validators for `TestStep`

### DIFF
--- a/scripts/pipeline_generator/step.py
+++ b/scripts/pipeline_generator/step.py
@@ -1,9 +1,72 @@
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, root_validator, model_validator
 from typing import List, Dict, Any, Optional
-
-from .utils import AgentQueue
+from typing_extensions import Self
+from .utils import AgentQueue, GPUType
 
 BUILD_STEP_KEY = "build"
+DEFAULT_TEST_WORKING_DIR = "/vllm-workspace/tests"
+
+class TestStep(BaseModel):
+    """This class represents a test step defined in the test configuration file."""
+    label: str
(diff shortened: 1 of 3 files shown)


scripts/pipeline_generator/step.py
```diff
@@ -1,9 +1,72 @@
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, root_validator, model_validator
 from typing import List, Dict, Any, Optional
-
-from .utils import AgentQueue
+from typing_extensions import Self
+from .utils import AgentQueue, GPUType
```
aslonnie, line 4: keep the separate line between local and third-party imports?

scripts/pipeline_generator/step.py
```diff
@@ -1,9 +1,72 @@
...
+            raise ValueError("'num_gpus' must be defined if 'num_nodes' is defined.")
+        if self.num_nodes and len(self.commands) != self.num_nodes:
+            raise ValueError("Number of commands must match the number of nodes.")
+        return self
+
+
+class BuildkiteStep(BaseModel):
+    """This class represents a step in Buildkite format."""
+    label: str
+    key: str
```
aslonnie, line 60: `key` is always required?

aslonnie (approved):

scripts/pipeline_generator/step.py
khluu: done

scripts/pipeline_generator/step.py
khluu: oh `key` is not required

khluu: I made it optional now

PR #40, opened 2024-10-07
[4/n][pipeline-gen] Define PipelineGeneratorConfig

khluu:
Define `PipelineGeneratorConfig` class to use for `PipelineGenerator`

### DIFF
--- a/scripts/pipeline_generator/pipeline_generator.py
+++ b/scripts/pipeline_generator/pipeline_generator.py
@@ -1,6 +1,6 @@
 import os
 import re
-from typing import List
+from typing import List, Optional
 
 from pydantic import BaseModel, field_validator
 
@@ -15,7 +15,7 @@ class PipelineGeneratorConfig:
         external_hardware_test_path: str,  # List of external hardware tests
         pipeline_file_path: str,  # Path to the output pipeline file
         run_all: bool = False,
-        list_file_diff: List[str] = []
+        list_file_diff: Optional[List[str]] = None,
     ):
         self.run_all = run_all
         self.list_file_diff = list_file_diff
--- a/scripts/pipeline_generator/pipeline_generator.py
+++ b/scripts/pipeline_generator/pipeline_generator.py
@@ -1,0 +1,50 @@
+import os
+import re
+from typing import List
+
+from pydantic import BaseModel, field_validator
+
+class PipelineGeneratorConfig(BaseModel):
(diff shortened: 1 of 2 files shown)


scripts/pipeline_generator/pipeline_generator.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @property
+    def container_image(self) -> str:
+        return f"{self.container_registry}/{self.container_registry_repo}:{self.commit}"
+
+    @field_validator("test_path")
+    @classmethod
+    def check_test_path(cls, value: str) -> str:
+        if not os.path.exists(value):
+            raise ValueError(f"Test path {value} does not exist")
```
aslonnie, line 26: field validator is supposed to be checking if the values of the fields are "legal", and they are supposed to be pure functions. it is part of the marshal / unmarshaling operation of a data structure, so it cannot have dependencies on the file system.

for example, if one marshals an object into bytes, it should always be able to unmarshal the bytes back to the object, no matter what the state of the file system is (or if it even has a file system).

please check against the file system in another explicit function.

khluu: I removed these validators for now.. will add them into helper later on

scripts/tests/pipeline_generator/test_pipeline_generator.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+        assert pipeline_generator_config.container_image == f"{container_registry}/{container_registry_repo}:{TEST_COMMIT}"
+
+
+def test_get_pipeline_generator_config_invalid_commit():
+    with tempfile.TemporaryDirectory() as temp_dir:
+        with open(os.path.join(temp_dir, TEST_FILE_PATH), "w") as f:
+            f.write("content")
+        with open(os.path.join(temp_dir, EXTERNAL_HARDWARE_TEST_FILE_PATH), "w") as f:
+            f.write("content")
```
aslonnie, line 40: you do not need these test file setup now?

khluu: removed

aslonnie (commented):
discussed offline. please drop pydantic as this is an internal object rather than for marshaling.

khluu:
Updated with built in Python class and separate `validate` method

aslonnie (approved):

PR #41, opened 2024-10-08
[5/n][pipeline-gen] Helper method to determine whether test should automatically run

khluu:
Determine if test should automatically run based on these logic:
- If specified as optional, test never run automatically (even when `run_all` is true)
- If there is no source file dependencies, test runs
- If `run_all` is specified, test runs
- If one of the files in `list_file_diff` match with source file dependencies, test runs

### DIFF
--- a/scripts/pipeline_generator/pipeline_generator_helper.py
+++ b/scripts/pipeline_generator/pipeline_generator_helper.py
@@ -1,0 +1,13 @@
+from typing import List
+
+from .step import TestStep
+
+def step_should_run(step: TestStep, run_all: bool, list_file_diff: List[str]) -> bool:
+    """Determine whether the step should automatically run or not."""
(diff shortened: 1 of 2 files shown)


aslonnie (commented):
do you have a PR that has everything together that I can have a look first?

khluu:
> do you have a PR that has everything together that I can have a look first?

This is the PR with everything together: https://github.com/vllm-project/buildkite-ci/pull/36

aslonnie (commented):
talked offline. next step:

- could we have a basic setup of running the high-level thing e2e? like with an example dir with the input files and make it clear that these files are the input files?
- and then, maybe in the same PR, have code that parses the input files into `TestStep`?

by having that first, we set up the context for the test step filtering logic in this PR.

PR #42, opened 2024-10-08
[6/n][pipeline-gen] Helper method to get build commands

khluu:
To not overcomplicate the main class with a bunch of hard-coded commands, this is moved to a helper function..


aslonnie (commented):
same as 5/n, PR with high-level context first please.

PR #43, opened 2024-10-08
[7/n][pipeline-gen] Helper method to get plugin config

khluu:
- Helper method to call relevant methods to grab plugin config depends on the GPU type.
- Currently, if GPU is A100, K8s plugin would be used. Otherwise, Docker plugin is used.


aslonnie (commented):
same as 5/n, PR with high-level context first please.

without examples/context of what `TestStep` looks like in a CI pipeline's context, it is not clear how to read the code in this PR.

PR #44, opened 2024-10-09
[5/n][pipeline-gen] Method to read test steps from Yaml file

khluu:


scripts/pipeline_generator/pipeline_generator.py
```diff
@@ -53,3 +52,24 @@ def __init__(
...
+
+def read_test_steps(file_path: str) -> List[TestStep]:
+    """Read test steps from test pipeline yaml and parse them into TestStep objects."""
+    with open(file_path, "r") as f:
+        content = yaml.safe_load(f)
+    return [TestStep(**step) for step in content["steps"]]
+
+@click.command()
+@click.option("--test_path", type=str, required=True, help="Path to the test pipeline yaml file")
+@click.option("--run_all", type=str)
```
aslonnie, line 64: maybe add help string that says `"1"` means true? and all other values means false?

scripts/pipeline_generator/pipeline_generator.py
```diff
@@ -53,3 +52,24 @@ def __init__(
...
+@click.option("--list_file_diff", type=str)
```
aslonnie, line 65: is this a file or a diff? could you add a help for this too?

aslonnie (approved):

scripts/pipeline_generator/pipeline_generator.py
khluu: done

scripts/pipeline_generator/pipeline_generator.py
khluu: done

PR #45, opened 2024-10-09
[7/n][pipeline-gen] Method to write Buildkite steps to yaml file

khluu:


scripts/tests/pipeline_generator/test_pipeline_generator.py
```diff
@@ -60,5 +62,25 @@ def test_read_test_steps():
...
+        BuildkiteStep(label="Test 1", commands=['echo "Test1.1"', 'echo "Test1.2"']),
+        BuildkiteStep(label="Test 2", commands=["command3"], agents = {"queue": AgentQueue.AWS_1xL4.value}),
+        BuildkiteBlockStep(block="Run Test 3", key="block-test-3"),
+        BuildkiteStep(label="Test 3", commands=["command4"], depends_on="block-test-3"),
+    ]
+    with tempfile.TemporaryDirectory() as temp_dir:
+        output_file_path = os.path.join(temp_dir, "output.yaml")
+        write_buildkite_steps(steps, output_file_path)
+        with open(output_file_path, "r") as f:
+            output = yaml.safe_load(f)
```
aslonnie, line 81: why `safe_load` here? I think this should just compare the raw content of the file without yaml parsing?

aslonnie (approved):

scripts/tests/pipeline_generator/test_pipeline_generator.py
khluu: Because the order of the keys can be different but after all they are the same.. anyway I updated it to compare the raw content instead

PR #46, opened 2024-10-09
[6/n][pipeline-gen] Update buildkite step format and add validators

khluu:
- Change type from `AgentQueue` to `str` so it can be dumped into yaml file 
- Add validator to make sure agent queue is valid


aslonnie (approved):
please update the PR description.

PR #47, opened 2024-10-15
Add python-only development tests

cermeng:
Add trigger for the test introduced in https://github.com/vllm-project/vllm/pull/9370.

### DIFF
--- a/scripts/test-template-fastcheck.j2
+++ b/scripts/test-template-fastcheck.j2
@@ -32,6 +32,13 @@ steps:
     command: bash .buildkite/run-neuron-test.sh
     soft_fail: false
 
+  - label: "Python-only Development Test"
+    depends_on: image-build
+    agents:
+      queue: small_cpu_queue
+    command: bash .buildkite/run-code-dev-test.sh
+    soft_fail: true
+
   {% for step in steps %}
   {% if step.gpu != "a100" and step.fast_check == true and step.num_nodes < 2 %}
   - label: "{{ step.label }}"


khluu:
Does this test only run on-demand?

cermeng:
> Does this test only run on-demand? 

You mean not to test it in fastcheck？ It takes less than 1 minute on my machine if vllm docker cache exists, so I put it in fastcheck. However, this test is not important and I think related scripts will not change frequently, so it's also reasonable not to include it in fast check

khluu:
> > Does this test only run on-demand?
> 
> You mean not to test it in fastcheck？ It takes less than 1 minute on my machine if vllm docker cache exists, so I put it in fastcheck. However, this test is not important and I think related scripts will not change frequently, so it's also reasonable not to include it in fast check

Oh what I meant is does it need to run for every commit or just whenever the PR author needs & unblocks it. By the way the test should be added here: https://github.com/vllm-project/vllm/blob/main/.buildkite/test-pipeline.yaml (you can add `fast_check: true` if you want it to run on just fastcheck)

cermeng:
> > > Does this test only run on-demand?
> 
> > 
> 
> > You mean not to test it in fastcheck？ It takes less than 1 minute on my machine if vllm docker cache exists, so I put it in fastcheck. However, this test is not important and I think related scripts will not change frequently, so it's also reasonable not to include it in fast check
> 
> 
> 
> Oh what I meant is does it need to run for every commit or just whenever the PR author needs & unblocks it. By the way the test should be added here: https://github.com/vllm-project/vllm/blob/main/.buildkite/test-pipeline.yaml (you can add `fast_check: true` if you want it to run on fastcheck or `fast_check_only: true` if it only runs on fastcheck )  

My previous reply can still answer this question: it is just a less important and time-consuming test, and I think there is no need to run for every commit

> this test is not important and I think related scripts will not change frequently

Why I add the test here is that I found all fast check are based on docker that has already installed vllm(dockerfile test target). If you can take a quick look, this test in the pr runs in a clean docker( I choose the base target in the dockerfile). I'm not so familiar with cool CI/CD, correct me if I'm wrong

cermeng:
Refer to https://github.com/vllm-project/vllm/pull/9370#issuecomment-2418159031 We won’t need to add task here after reconstruction. Close it.

PR #48, opened 2024-10-22
[8/n][pipeline-gen] Add method to generate build step

khluu:
- Add method to generate build step based on commit, container registry and image.

### DIFF
--- a/scripts/pipeline_generator/pipeline_generator_helper.py
+++ b/scripts/pipeline_generator/pipeline_generator_helper.py
@@ -1,0 +1,26 @@
+from typing import List
+
+def get_build_commands(container_registry: str, buildkite_commit: str, container_image: str) -> List[str]:
+    ecr_login_command = (
+        "aws ecr-public get-login-password --region us-east-1 | "
+        f"docker login --username AWS --password-stdin {container_registry}"
+    )
+    image_check_command = f"""#!/bin/bash
+if [[ -z $(docker manifest inspect {container_image}) ]]; then
+echo "Image not found, proceeding with build..."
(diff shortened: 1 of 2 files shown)


aslonnie:
is this for review? could you add a PR description maybe?

scripts/pipeline_generator/pipeline_generator_helper.py
```diff
@@ -0,0 +1,26 @@
+from typing import List
+
+def get_build_commands(container_registry: str, buildkite_commit: str, container_image: str) -> List[str]:
```
aslonnie, line 3: any unit test? or is this so simple that does not require one?

maybe we should have some form of integration test to test these commands.

khluu: It's just replacing placeholders for container registry, container image, and commit so I don't think unit test is needed here.

scripts/pipeline_generator/pipeline_generator_helper.py
```diff
@@ -0,0 +1,26 @@
...
+fi
+"""
+    docker_build_command = (
+        f"docker build "
+        f"--build-arg max_jobs=64 "
+        f"--build-arg buildkite_commit={buildkite_commit} "
+        f"--build-arg USE_SCCACHE=1 "
+        f"--tag {container_image} "
+        f"--target test "
+        f"--progress plain ."
```
aslonnie, line 23: could you leave a todo here to not using `.` to include everything as the build context?

aslonnie (approved):

PR #50, opened 2024-10-24
[10/n][pipeline-gen] Add helper method to convert test step into Buildkite step

khluu:
- Add a method to get plugin config upon whether GPU type is provided and which type of GPU is requested. Since A100 GPU is on EKS, all tests with A100 GPU would be using K8s plugin and all other tests run with Docker plugin.
- Add a method to convert `TestStep` into `BuildkiteStep` based on whether the test is a regular one or a multi node one (since multi-node test uses a custom script to run)
- The helper method will be used in `PipelineGenerator` where the test steps are processed into `BuildkiteStep` and `BuildkiteBlockStep`. The next PR will add the method since it contains some logic that is better to put in a separate PR. The logic is here: https://github.com/vllm-project/buildkite-ci/pull/36/files#diff-82edd9d64d3e791b9faade2e3aff441908121926000ca52bd54a8f17c976f4d0R120

### DIFF
--- a/scripts/pipeline_generator/plugin.py
+++ b/scripts/pipeline_generator/plugin.py
@@ -54,6 +54,7 @@ class DockerPluginConfig(BaseModel):
     command: List[str] = Field(default_factory=list)
     environment: List[str] = DEFAULT_DOCKER_ENVIRONMENT_VARIBLES
     volumes: List[str] = DEFAULT_DOCKER_VOLUMES
+    shell: List[str] = ["/bin/bash", "-c"]
 
 
 class KubernetesPodContainerConfig(BaseModel):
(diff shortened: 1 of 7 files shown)


scripts/pipeline_generator/plugin.py
```diff
@@ -54,6 +54,7 @@ class DockerPluginConfig(BaseModel):
     command: List[str] = Field(default_factory=list)
     environment: List[str] = DEFAULT_DOCKER_ENVIRONMENT_VARIBLES
     volumes: List[str] = DEFAULT_DOCKER_VOLUMES
+    shell: List[str] = ["/bin/bash", "-c"]
```
khluu, line 57: This file contains changes from https://github.com/vllm-project/buildkite-ci/pull/49/files which would be rebased after that PR is merged to main

aslonnie (approved):

PR #78, opened 2025-03-28
Enable the AMD testing for a passing subset of tests.

Alexei-V-Ivanov-AMD:
Enable the AMD testing for a passing subset of tests.


andy-neuma (approved):
yeah, good to be more strict. thanks!

PR #85, opened 2025-04-04
Added slack notification for Power job failures

npanpaliya:


scripts/test-template-ci.j2
```diff
@@ -324,6 +324,26 @@ steps:
...
+         cat <<- YAML | buildkite-agent pipeline upload
+         steps:
+           - label: "Notify owners about failing test"
+             agents:
+               queue: ibm-ppc64le
+             command: echo "IBM Power(ppc64le) Build/Test failed"
+             notify:
+               - slack:
+                   channels:
+                     - "vLLM#vllm-ci-on-power"
```
khluu, line 343: this is just `#vllm-ci-on-power`

khluu: I think so https://github.com/vllm-project/ci-infra/blob/e33e25db9bd4836cdc0c8c9dda9de64d60a3d10a/scripts/test-template-fastcheck.j2#L284

khluu (approved):

PR #87, opened 2025-04-21
add torch nightly in ci

yangw-dev:


scripts/test-template-ci.j2
```diff
@@ -243,9 +245,127 @@ steps:
   {% endif %}
   {% endfor %}
 
+- group: "vllm against torch nightly"
```
huydhn, line 248: Just curious if we can use the `matrix` parameter here to run stable and nightly side by side instead of copy/paste the CI steps.  It seems doable if we could set `soft_fail` dynamically.  What do you think?

yangw-dev: hmm, I wonder if it's possible,so we only want  the nightlyto be triggered manually only, let me try it to see if it's possible

yangw-dev: add macro functions to wrap configs that can be used for both vllm against torch nightly and normal vllm.

Notice, kevin is working on use python script to generate the yaml file for test, so this jinja template thing will be deprecated in h2

PR #93, opened 2025-04-25
[Enable nightly run for vllm against torch nightly

yangw-dev:
fix the issue created by removing the macros:

Place block for normal runs,but make the  build step`vllm against torch nightly` unblock for nightly run
Testing:
NIGHTLY=1
https://buildkite.com/vllm/ci/builds/18615

NIGHTLY=0
https://buildkite.com/vllm/ci/builds/18613#01966e29-17cd-4c8b-bf23-94eed149d295

### DIFF
--- a/buildkite/test-template-ci.j2
+++ b/buildkite/test-template-ci.j2
@@ -299,15 +299,31 @@ steps:
               limit: 2
     {% for step in steps %}
     {% if step.torch_nightly %}
+      {% set ns = namespace(blocked=1) %}
+      {% if nightly == "1" %}
+      {% set ns.blocked = 0 %}
+      {% endif %}
 
-      {% if step.optional and nightly != "1" %}
+      {% if step.source_file_dependencies %}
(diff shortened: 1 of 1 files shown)


buildkite/test-template-ci.j2
```diff
@@ -294,15 +299,31 @@ steps:
               limit: 2
     {% for step in steps %}
     {% if step.torch_nightly %}
+      {% set ns = namespace(blocked=1) %}
+      {% if nightly == "1" %}
+      {% set ns.blocked = 0 %}
+      {% endif %}
 
-      {% if step.optional and nightly != "1" %}
```
yangw-dev, line 298: we want to only trigger certain path of the tests too, also i think previous it misses:
{% if nightly == "1" %}
   {% set ns.blocked = 0 %}
{% endif %}

so this may make the nightly run not trigger properly

khluu (approved):

PR #116, opened 2025-06-27
[HPU][NOTIFY] update hpu CI with notification

xuechendi:
We want to add a notification in case of failing


xuechendi:
@khluu  @DarkLight1337 , the PR seems still need one approval from maintainer, may you help me to check
The goal of this PR is to send notification to collab-intel-ci channel for HPU test failing

PR #119, opened 2025-07-02
Reassign AMD testing queues

Alexei-V-Ivanov-AMD:
Update test-template-ci.j2


seemethere:
Howdy! Mostly LGTM! Can we run a test and link it in the description?

seemethere:
I was mostly asking if we've ran with this particular template in CI. Worse thing would be a template rendering issue so mostly looking to validate that this renders correctly

seemethere (approved):

PR #143, opened 2025-08-25
[experimental] pull prebuilt test image + mount code when we can instead.

dougbtv:
- Skip Docker builds when VLLM_USE_PRECOMPILED=1 and no critical changes detected.
- Pull branch-appropriate prebuilt image (:latest) and bind-mount workspace at /vllm-workspace.
- Bootstrap inside container: upgrade pip/uv; if requirements changed, install common/build/test; always `uv pip install --system -e .`.
- Guard all image-build steps behind `skip_image_build`; switch `depends_on` accordingly.
- Respect `SKIP_IMAGE_BUILD` env override and log final decision + chosen image.

PR #149, opened 2025-08-28
[HOLD] CI: pull prebuilt test image and overlay code when appropriate to do so

dougbtv:
- Skip Docker builds when VLLM_USE_PRECOMPILED=1 and no critical changes detected.
- Pull prebuilt image from post-commit on main (:latest) and copy code in /vllm-workspace.
- Bootstrap inside container: upgrade pip/uv; if requirements changed, install common/build/test;
- Uses a macro for re-used stanzas to not make the docker steps look horrific

Replaces #143 

Test job with same code before squash: https://buildkite.com/vllm/ci/builds/28758

I want to hold merging this for now, it's pending some more testing, things I need to check on...

* Is it appropriately running for all involved tests?
* There's this logic for determining that an image is pushed for a given commit.... I have to look into it, this means, some won't push during CI runs and only post-commit on main? I have to check it.


dougbtv:
Most recent test run: https://buildkite.com/vllm/ci/builds/28845

Running by doing "new build", then:

Commit: HEAD
Branch: khluu/test_fixed_premerge (main was causing too much noise)
ENV vars:

```
BUILDKITE_PULL_REQUEST=23771
VLLM_CI_BRANCH=feature/docker-checkout-overlay
```

PR #153, opened 2025-09-03
Add default cancel_on_build_failing

tmuttaki:


tmuttaki:
Nvm, not making this change now.

PR #154, opened 2025-09-04
Fast Fail Test Groups

tmuttaki:
Setting `PYTEST_ADDOPTS=-x` on the pipeline
Setting -e flag for bash to early exit

Example run with fast fail: https://buildkite.com/vllm/fastcheck/builds/39726#019914e1-5975-4b7b-9c2c-79f467ad2e1a


dougbtv (approved):
lgtm! makes sense to have bash exit on error, and I take it the -x for python is so it doesn't miss trapping an exit code that's from another subprocess or something? at any rate, looks good :+1:

PR #171, opened 2025-09-23
Skip image build when eligible

dougbtv:
In this iteration, I'd like to figure out how to test with the commits on main, too.


dougbtv:
latest build after rebase (and refactor to some extent to integrate with code coverage command injection): https://buildkite.com/vllm/ci/builds/33889

PR #177, opened 2025-09-29
Skip hardware tests on docs-only changes

rzabarazesh:
This will skip hardware CI tests on docs only PRs such as https://github.com/vllm-project/vllm/pull/25819 

Testing
---
Normal run on master:
https://buildkite.com/vllm/ci/builds/32829/steps/canvas

PR with just docs files modified:
https://buildkite.com/vllm/ci/builds/32831/steps/canvas

PR #178, opened 2025-09-29
skip pipeline if all changed files are under docs/

hl475:
we could skip the entire pipleine from bootstrap stage if the changes only touching docs


hl475:
@khluu @yeqcharlotte @seemethere

buildkite/bootstrap.sh
```diff
@@ -105,6 +109,40 @@ if [[ $BUILDKITE_BRANCH == "main" ]]; then
     file_diff=$(get_diff_main)
 fi
 
+# ----------------------------------------------------------------------
+# Early exit start: skip pipeline if conditions are met
+# ----------------------------------------------------------------------
+
+# skip pipeline if all changed files are under docs/
+if [[ "${DOCS_ONLY_DISABLE}" != "1" && "${RUN_ALL}" != "1" && "${NIGHTLY}" != "1" ]]; then
```
khluu, line 117: can we just do `DOCS_ONLY` here?

khluu (commented):
Thanks! Can you test using a PR with only docs changes to see if it works?

buildkite/bootstrap.sh
hl475: absolutely! just updated the PR, thanks!

hmellor:
Is this different to https://github.com/vllm-project/ci-infra/pull/177? Should the older PR be reverted if this PR is preferred?

hl475:
@khluu per instructions from https://github.com/vllm-project/ci-infra?tab=readme-ov-file#how-to-test-changes-in-this-repo, could you please help me add into the repo, so I can create a feature branch? Thanks!

hmellor:
I think prefer this solution as it also skips the image build

hl475:
will close this PR and use https://github.com/vllm-project/ci-infra/pull/180 instead

PR #183, opened 2025-10-01
Remove testing on MI250

Alexei-V-Ivanov-AMD:
This PR removes the AMD tests that are supposed to be running on MI250.


seemethere:
Did we run a test workflow for this?

seemethere (approved):
LGTM

PR #338, opened 2026-04-15
[CI] Support custom Buildkite agent tags in pipeline generator

zxd1997066:
This PR adds support for custom tags in the pipeline generator.

Previously, generated steps only set agents.queue from device-based mapping. ```agents={"queue": get_agent_queue(step)}```
With this change, a step can define additional agent constraints and they will be merged into the generated agents map.


khluu:
We also keep the logic to route to agents queue based on mapping for security reason. The job specs from upstream cannot define agent queue

PR #347, opened 2026-04-23
[ROCm][CI] Route AMD-only Buildkite steps through AMD queues and consolidate AMD pipeline

AndreasKaratzas:
This PR updates the Python Buildkite pipeline generator so ROCm tests can be defined as AMD-only steps directly in the vLLM test-area YAML files.

### Summary

- Recognize top-level AMD GPU devices such as `mi250_*`, `mi300_*`, `mi325_*`, and `mi355_*`.
- Render those steps as AMD hardware jobs with:
  - `AMD:` labels
  - matching `amd_mi*` Buildkite queues
  - `image-build-amd` dependencies
  - `bash .buildkite/scripts/hardware_ci/run-amd-test.sh`
  - `VLLM_TEST_COMMANDS` passed through the environment
- Use AMD setup commands for AMD direct and mirror steps:
  - include `amd-smi`
  - exclude NVIDIA setup and CUDA coredump exports
- Add a single automatic retry rule for common transient AMD/agent failures.
- Tighten source dependency matching to exact file or directory-prefix matches.
- Keep AMD mirror gating aligned with the same auto-run/manual-block behavior.
- Refresh the pipeline generator tests and add coverage for direct AMD test-area steps.

### Motivation

This enables vLLM PRs to add AMD-only jobs in `.buildkite/test_areas/*.yaml` with entries like `device: mi300_4` or `device: mi300_8`.

Previously, those direct AMD test-area steps could fall through to the default NVIDIA GPU queue path, which meant they appeared without the `AMD:` prefix and could run on `gpu_1_queue`. This fixes that path so AMD-only tests route to the intended AMD hardware.

This is needed for `vllm-project/vllm#45481`, where the new LM Eval jobs should run on MI300 queues.

### Generated Behavior

For a direct AMD test-area step like:

```yaml
- label: LM Eval MiniMax M2-7 (4xMI300)
  key: lm-eval-minimax-m27-mi300
  depends_on:
    - image-build-amd
  device: mi300_4
  optional: true
  soft_fail: true
```

the generator now emits a Buildkite command step with:

- label: `AMD: LM Eval MiniMax M2-7 (4xMI300) (mi300_4)`
- queue: `amd_mi300_4`
- command: `bash .buildkite/scripts/hardware_ci/run-amd-test.sh`
- dependency: `image-build-amd`
- env: `VLLM_TEST_COMMANDS=...`
- retry: one automatic retry for `exit_status` `[-1, 1, 128]` and `signal_reason` `["none", "agent_stop", "agent_refused"]`


buildkite/pipeline_generator/utils_lib/git_utils.py
```diff
@@ -25,32 +25,29 @@ def get_merge_base_commit() -> Optional[str]:
...
             output = subprocess.check_output(
-                ["git", "diff", "--name-only", "--diff-filter=ACMDR", "HEAD~1"],
+                ["git", "diff", "--name-only", "--diff-filter=ACMDR", "HEAD~1", "HEAD"],
                 universal_newlines=True,
             )
         else:
             merge_base = merge_base_commit
             if not merge_base:
-                pass
+                raise RuntimeError("Failed to determine merge base commit for git diff.")
```
AndreasKaratzas: @khluu Lmk if I should revert this part. I added it cause I thought that we should not silently pass right?

buildkite/pipeline_generator/buildkite_step.py
```diff
@@ -144,15 +147,34 @@ def _get_variables_to_inject() -> Dict[str, str]:
...
+            "export CUDA_ENABLE_COREDUMP_ON_EXCEPTION=1 && export CUDA_COREDUMP_SHOW_PROGRESS=1 && export CUDA_COREDUMP_GENERATION_FLAGS='skip_nonrelocated_elf_images,skip_global_memory,skip_shared_memory,skip_local_memory,skip_constbank_memory'",
+        ]
+
+    if setup_profile == "amd":
+        return [
+            "echo '--- :amd: GPU Info'",
+            "(command amd-smi || true)",
+        ]
+
+    raise ValueError(f"Unsupported setup profile: {setup_profile}")
```
AndreasKaratzas, line 387: Added different profile for AMD and that also gives flexibility for anyone else.

buildkite/pipeline_generator/buildkite_step.py
```diff
@@ -205,6 +227,20 @@ def _create_block_step(step: Step, list_file_diff: List[str]) -> BuildkiteBlockS
...
+    return diff_file == normalized or diff_file.startswith(f"{normalized}/")
+
+
+def _step_is_blocked(step: Step, list_file_diff: List[str]) -> bool:
+    global_config = get_global_config()
+    return (not _step_should_run(step, list_file_diff) or (
+        step.optional and global_config["nightly"] != "1"
+    ))
+
+
```
AndreasKaratzas: The generator used raw substring matching:

`if source_file in diff_file`
That is too loose. Examples:

`vllm/` would match `notvllm/foo.py`
`src/file` would match `src/file_extra.py`
The helper changes that to proper exact-or-prefix path matching:

exact file match: `diff_file == normalized`
directory prefix match: `diff_file.startswith(f"{normalized}/")`
So this one exists to make gating correct.

`_step_is_blocked()` is mostly a cleanup/helper.

I added it because the “should this step be behind a manual block?” logic was being repeated.

AndreasKaratzas (commented):
@khluu These changes affect the upstream CI logic too, so check them out if you like them or if I should remove them.

buildkite/pipeline_generator/utils_lib/git_utils.py
```diff
@@ -25,32 +25,29 @@ def get_merge_base_commit() -> Optional[str]:
 def get_list_file_diff(branch: str, merge_base_commit: Optional[str]) -> List[str]:
     """Get list of file paths that get changed between current branch and origin/main."""
     try:
-        subprocess.run(["git", "add", "."], check=True)
         if branch == "main":
             output = subprocess.check_output(
-                ["git", "diff", "--name-only", "--diff-filter=ACMDR", "HEAD~1"],
+                ["git", "diff", "--name-only", "--diff-filter=ACMDR", "HEAD~1", "HEAD"],
```
khluu: why are we adding `HEAD` at the end?

khluu (commented):
can you also break this down to smaller PRs?

buildkite/pipeline_generator/utils_lib/git_utils.py
AndreasKaratzas: i removed it .. i also simplified this pr to basics towards consolidation of amd ci. rest of them can either wait or not be applied at all, they are too optional.

AndreasKaratzas:
Example run: https://buildkite.com/vllm/ci/builds/73140

AndreasKaratzas:
The changes in the infra side are passing (I am just going to remove the Kimi test group until the MI355 cluster is ready for it.

AndreasKaratzas:
AMD full CI run with this PR: https://buildkite.com/vllm/amd-ci/builds/9757/canvas
Upstream CI (test subset) run with this PR (against https://github.com/vllm-project/vllm/pull/45481): https://buildkite.com/vllm/ci/builds/73304/canvas

Harry-Chen (approved):
LGTM in principle and from CI results

buildkite/pipeline_generator/buildkite_step.py
```diff
@@ -146,6 +189,7 @@ def to_yaml(self):
             "env": self.env,
             "parallelism": self.parallelism,
             "priority": self.priority,
+            "timeout_in_minutes": self.timeout_in_minutes,
```
khluu: let's not add this for now

khluu (commented):
Thanks for this! The `SetupProfile` / `_get_setup_commands` split is the right direction — parameterize the NVIDIA/AMD difference instead of forking. The rest of the AMD support, though, adds a parallel `_amd_*` path that re-implements logic the NVIDIA path already has. A few consolidations would make it cleaner and more reusable across NVIDIA/AMD:

**1. Merge the two AMD step constructors.** `_create_amd_direct_step` and `_create_amd_mirror_step` build the same `BuildkiteCommandStep` — same `run-amd-test.sh` command, `_get_amd_label`, `_get_amd_queue`, `_get_amd_env`, `priority=200`, `AMD_RETRY`, `parallelism`, `timeout_in_minutes`. Only the inputs differ. Collapse into one `_create_amd_step(*, label, device, commands_str, depends_on, extra_env, soft_fail, timeout_in_minutes, parallelism)` and let the two callers compute the inputs.

**2. Don't add a third device→queue map.** `AMD_GPU_QUEUE_MAP` + `_get_amd_queue` exist only because the AMD path bypasses `get_agent_queue`, which is already the single place that maps device→queue. It also collides with #317 (consolidating device→queue into constants.py). Fold the AMD GPU devices into `get_agent_queue` / the constants map. The mapping is mechanical (`AgentQueue.value == "amd_" + DeviceType.value` for every entry), so it can be derived rather than hand-listed.

**3. Unify the "should run" gating.** `_amd_mirror_should_run` re-implements `_step_should_run` plus AMD overrides, and the new `_source_file_dependencies_match` is only used by the former — `_step_should_run` still has its own inline loop. Make `_step_should_run` use `_source_file_dependencies_match`, then express the mirror case as the shared function over an effective step: `step.model_copy(update={"optional": ..., "source_file_dependencies": ...})` (the mirror path already uses `model_copy` for commands). Also note the direct AMD path uses `_step_should_run` while the mirror path uses `_amd_mirror_should_run`, so the two AMD entry points gate differently today — worth making deliberate.

**4. Block-step creation is now written 3×** (NVIDIA inline, AMD-direct, AMD-mirror), with subtly different keys. Extract one small helper that builds the block step and stitches its key into the command step's `depends_on`.

Minor:
- Behavior change: mirror steps go from `retry=None` → `AMD_RETRY` (one auto-retry). Intended per the description, just flagging.
- `_get_amd_container_timeout_s`: the `-60` ("time out the container a minute before Buildkite") could use a one-line comment.

Net: the `SetupProfile` idea is good — the items above are about finishing that same "one parameterized path" approach instead of branching around it.

AndreasKaratzas:
AMD full CI run with this PR:https://buildkite.com/vllm/amd-ci/builds/9800/canvas 
Upstream CI (test subset) run with this PR (against [vllm-project/vllm#45481](https://github.com/vllm-project/vllm/pull/45481)): https://buildkite.com/vllm/ci/builds/73459/canvas

PR #355, opened 2026-05-12
Disable expandable_segments for h200_18gb MIG docker plugin

khluu:
## Summary
- Add `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False` to the h200_18gb docker plugin environment
- This fixes a PyTorch NVML assertion failure that prevents most vLLM tests from running on H200 18GB MIG partitions

## Problem
When running vLLM CI jobs on H200 18GB MIG partitions, PyTorch's `CUDACachingAllocator` crashes with:
```
RuntimeError: NVML_SUCCESS == r INTERNAL ASSERT FAILED at
"/pytorch/c10/cuda/CUDACachingAllocator.cpp":1165
```

The expandable segments feature (default in PyTorch 2.11) uses NVML to query device memory info. On MIG partitions, these NVML calls to the parent GPU fail with permission errors, causing the assertion.

## Fix
Setting `expandable_segments:False` forces PyTorch to use traditional `cudaMalloc` instead of the virtual memory + NVML path. This is safe for single-GPU MIG workloads.

## Validation
Tested in Buildkite build [#65789](https://buildkite.com/vllm/ci/builds/65789) — this was the root cause for 45+ out of 53 job failures when migrating `gpu_1_queue` jobs to h200_18gb MIG.

### DIFF
--- a/buildkite/pipeline_generator/plugin/docker_plugin.py
+++ b/buildkite/pipeline_generator/plugin/docker_plugin.py
@@ -29,6 +29,7 @@ h200_18gb_plugin_template = {
     "environment": [
         "VLLM_USAGE_SOURCE=ci-test",
         "NCCL_CUMEM_HOST_ENABLE=0",
+        "PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False",
         "HF_TOKEN",
         "HF_HOME",
         "CODECOV_TOKEN",
--- a/buildkite/pipeline_generator/plugin/docker_plugin.py
+++ b/buildkite/pipeline_generator/plugin/docker_plugin.py
@@ -29,7 +29,7 @@ h200_18gb_plugin_template = {
     "environment": [
         "VLLM_USAGE_SOURCE=ci-test",
         "NCCL_CUMEM_HOST_ENABLE=0",
-        "PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False",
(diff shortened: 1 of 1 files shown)


khluu:
## Update: `expandable_segments:False` is a partial fix

Build [#65899](https://buildkite.com/vllm/ci/builds/65899) tested this fix. Results so far (21/66 completed):
- **3 test jobs passed** (Kernels KDA Test, Python-only Installation, Basic Models Tests Initialization)
- **16 test jobs failed** — the NVML assertion STILL fires in EngineCore subprocesses

The `expandable_segments:False` setting prevents the NVML call during initial PyTorch allocator setup, but **PyTorch's CUDACachingAllocator has additional NVML dependencies** that are triggered when allocating large tensors (KV cache) in vLLM's EngineCore subprocesses. The assertion at `CUDACachingAllocator.cpp:1165` is hit through a different code path than expandable segments.

### Possible next steps
1. Try `PYTORCH_CUDA_ALLOC_CONF=backend:cudaMallocAsync` — uses CUDA's native async allocator which doesn't use NVML at all (but may have compatibility issues with CUDA graphs)
2. Try adding `PYTORCH_NVML_BASED_CUDA_CHECK=0` in combination
3. Consider running MIG containers with `--cap-add=SYS_ADMIN` for full NVML access (security tradeoff)
4. Accept MIG limitations and only migrate pure kernel tests + lightweight init tests

PR #370, opened 2026-06-05
Update uLLM ci_monitoring startup to fix security risk

yiw-wang:


terraform/gcp_old/tpu-inference/cloud-ullm-inference-ci-cd/main.tf
```diff
@@ -128,7 +128,7 @@ module "ci_monitoring" {
   project_id            = var.project_id
   pipeline_slug         = "tpu-inference-ci"
   org_slug              = "tpu-commons"
-  buildkite_token_value = data.google_secret_manager_secret_version.buildkite_agent_token_ci_cluster.secret_data
+  buildkite_token_secret_id = data.google_secret_manager_secret_version.buildkite_agent_token_ci_cluster.secret
```
QiliangCui, line 131: there should be other folder i remember.

like test project that is under tpu-inference. could you check if that need to update as well? our v6e test machines are there.

QiliangCui (approved):

terraform/gcp_old/tpu-inference/cloud-ullm-inference-ci-cd/main.tf
yiw-wang: the original risk report ticket b/510376923 was just reporting how I use it to startup the buildkite-agent-metrics not buildkite-agent which used in other VMs, so we should be good to just change this one here.

QiliangCui (approved):

PR #374, opened 2026-06-12
Export TPU_VERSION and setup GCS Fuse on host VMs

theminghuang:
The benchmark script relies on TPU_VERSION to figure out the correct jax cache to download https://github.com/vllm-project/tpu-inference/blob/main/.buildkite/scripts/run_in_docker.sh#L125
We don't always set this from benchmark case (and we should not) and it results in downloading incorrect cache. Seeing v6e cache on v7 machine and all it does is occupies additional storages.

I'm also planning to enable a change to do the jax cache clean up after the docker run to save space on tpu, hopefully to reduce the times we need to do maintenance. As a result, we always download the cache from gcs and enabling  rapid cache  could greatly reduce the time. 

Screenshot 2026-06-11 at 11 20 26 PM


theminghuang:

- #374 ◀

QiliangCui:
@yiw-wang , could you take a look?

yiw-wang (approved):
changes LGTM, can you check if those format changes are needed?

theminghuang:
> changes LGTM, can you check if those format changes are needed?

not necessary but I usually do `terraform fmt` to format them automatically

QiliangCui (approved):

PR #377, opened 2026-06-15
Don't mount on the same path where we do rsync

theminghuang:


theminghuang:

- #377 ◀

terraform/gcp_old/tpu-inference/modules/ci_v5/main.tf
```diff
@@ -81,50 +81,18 @@ resource "google_tpu_v2_vm" "tpu_v5" {
...
-          ullm-ci-cache \
-          /mnt/disks/persist/tpu_jax_cache; then
-        
-        echo "GCS FUSE mount successful."
-        echo 'CI_CACHE_FUSE_MOUNTED=1' | sudo tee -a /etc/environment
-      else
-        echo "ERROR: Failed to mount GCS FUSE bucket."
-        exit 1
-      fi
+      echo 'CI_CACHE_FUSE_MOUNTED=1' | sudo tee -a /etc/environment
```
yiw-wang: if we are sure that the "mountpoint -q "$FUSE_MOUNT_DIR"" will be false, do we still need to force update `CI_CACHE_FUSE_MOUNTED=1` here

theminghuang: Yeah removed.

yiw-wang (approved):

PR #378, opened 2026-06-16
Run pre-commit as a dedicated step in parallel with image build

khluu:
## Summary

Follow-up to #312. That PR made the bootstrap step (pipeline generation) **block** on polling the pre-commit GitHub Actions check before generating/uploading the pipeline — so nothing, including the image build, started until pre-commit finished.

This PR splits pre-commit into a **dedicated Buildkite step** that runs **in parallel with the image build**, and makes every step that depends on an image build also depend on the pre-commit step. Tests still wait for pre-commit to pass, but pre-commit and the image build now run concurrently.

## Changes

- `buildkite_step.py`
  - `create_precommit_group_step(repo, commit)` — a `Pre-commit` group with one command step (`key: pre-commit`, `depends_on: []`, small CPU queue) that polls the `pre-commit` GitHub check-run and fails (with a Buildkite annotation) if it didn't pass / timed out (1500s, unchanged).
  - `add_precommit_dependency(...)` — appends `pre-commit` to the `depends_on` of every command step that depends on an image build, skipping the image-build steps themselves and block steps. Idempotent.
- `pipeline_generator.py` — removes the blocking `check_precommit_passed()` call; on PR builds, wires the dependency and appends the dedicated step.
- `utils_lib/git_utils.py` — removes the now-unused `check_precommit_passed()` and `time` import.

## Notes

- The poll command is pure stdlib Python (`urllib`/`json`), with the commit/repo baked in at generation time and **no `$`**, so `buildkite-agent pipeline upload` interpolation can't mangle it.
- Gating applies to any step depending on an `image-build*` key (main, CPU, AMD, arm64, torch-nightly), since pre-commit is a repo-wide gate.
- Trade-off: the image build now runs even when pre-commit ultimately fails (wasted work on failure), in exchange for parallelism in the common passing case.

## Testing

Triggered a vLLM Buildkite build with `VLLM_CI_BRANCH` pointing at this branch to verify pre-commit is a separate step and downstream jobs wait on it.

### DIFF
--- a/buildkite/pipeline_generator/buildkite_step.py
+++ b/buildkite/pipeline_generator/buildkite_step.py
@@ -39,11 +39,21 @@ while True:
         if conclusion == "success":
             print("pre-commit check passed on commit " + commit + ".")
             raise SystemExit(0)
-        subprocess.run(["buildkite-agent", "annotate", ":x: pre-commit check has not passed on this PR (conclusion: " + str(conclusion) + "). Please fix pre-commit issues before running CI.", "--style", "error"], check=False)
-        raise SystemExit("pre-commit check failed on commit " + commit + " (conclusion: " + str(conclusion) + ").")
+        # Only block downstream tests on a confirmed pre-commit failure. Other
+        # conclusions (skipped, cancelled, neutral, stale, ...) are not real
+        # failures, and pre-commit is also a required check on the PR, so let CI
+        # proceed rather than false-failing the whole build.
+        if conclusion in ("failure", "timed_out", "action_required"):
+            subprocess.run(["buildkite-agent", "annotate", ":x: pre-commit check failed on this PR (conclusion: " + str(conclusion) + "). Please fix pre-commit issues before running CI.", "--style", "error"], check=False)
+            raise SystemExit("pre-commit check failed on commit " + commit + " (conclusion: " + str(conclusion) + ").")
+        print("pre-commit check did not fail (conclusion: " + str(conclusion) + "); allowing CI to proceed.")
+        raise SystemExit(0)
     if elapsed >= max_wait:
-        subprocess.run(["buildkite-agent", "annotate", ":warning: Timed out waiting for pre-commit check to complete.", "--style", "warning"], check=False)
-        raise SystemExit("Timed out after " + str(max_wait) + "s waiting for pre-commit check on commit " + commit + ".")
+        # Do not block CI if the check never completes; the pre-commit check on
+        # the PR still surfaces a genuine failure independently.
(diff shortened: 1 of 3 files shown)


buildkite/pipeline_generator/buildkite_step.py
```diff
@@ -9,6 +9,92 @@
...
+    with urllib.request.urlopen(request) as response:
+        check_runs = json.load(response).get("check_runs", [])
+    precommit_run = next((run for run in check_runs if run["name"] == "pre-commit"), None)
+    if precommit_run and precommit_run["status"] == "completed":
+        conclusion = precommit_run["conclusion"]
+        if conclusion == "success":
+            print("pre-commit check passed on commit " + commit + ".")
+            raise SystemExit(0)
+        subprocess.run(["buildkite-agent", "annotate", ":x: pre-commit check has not passed on this PR (conclusion: " + str(conclusion) + "). Please fix pre-commit issues before running CI.", "--style", "error"], check=False)
+        raise SystemExit("pre-commit check failed on commit " + commit + " (conclusion: " + str(conclusion) + ").")
```
tahsintunan: _(Probably out of scope for this PR, but flagging it here since it's already moving this exact code)_

This carries over a false-fail introduced in #312: any non-`success` conclusion here hard-fails the build. Some examples:

- `skipped` → instant fail. [[Build 72406](https://buildkite.com/vllm/ci/builds/72406)]: the pre-commit job was skipped (anti-spam gate / superseded run), so the build died in 32s with `conclusion: skipped`.
- queued → timeout fail (lines 44-46). [[Build 72371](https://buildkite.com/vllm/ci/builds/72371)]: pre-commit sat `queued` the full 1500s and timed out.

Since this step only gates downstream tests now, could we fail only on a confirmed failure and let everything else through?

```python
if conclusion == "success":
  raise SystemExit(0)
if conclusion in {"failure", "timed_out", "action_required"}:
  ...annotate + fail...
raise SystemExit(0)  # skipped/cancelled/neutral: not a real failure
```

(and warn + pass on the timeout instead of failing). pre-commit also reports as its own check on the PR, so a genuine failure is still surfaced there independent of this step.

khluu: Good catch — agreed, this makes sense. Fixed in 7df1dd9.

The step now blocks downstream tests only on a confirmed failure (`failure`, `timed_out`, `action_required`). Everything else (`skipped`, `cancelled`, `neutral`, `stale`, `startup_failure`, …) passes through, and the poll timeout now warns + passes instead of failing. A genuine pre-commit failure is still surfaced via the pre-commit check on the PR, so we don't lose that signal.

PR #383, opened 2026-06-23
[ROCm] Fix retry matching and gate debug agent

AndreasKaratzas:
This fixes AMD CI automatic retries by splitting the retry policy into separate Buildkite rules, so normal `exit_status: 1` failures are no longer blocked by a missing `signal_reason`. It also makes the ROCm debug agent opt-in via `VLLM_CI_ENABLE_ROCM_DEBUG_AGENT=1`, since recent logs show it can crash inside `amd_dbgapi_event_processed` and mask the actual test behavior. By default, AMD steps now print that the debug agent is disabled instead of exporting `HSA_TOOLS_LIB` and `HSA_ENABLE_DEBUG`. Updated pipeline generator tests cover the new retry shape and both debug-agent disabled/enabled paths.


AndreasKaratzas:
Upstream CI validation run: https://buildkite.com/vllm/ci/builds/73784/canvas

PR #387, opened 2026-06-23
fix(buildkite): download codecov to secure temp dir to prevent bypass

jcc-google:
Always download the official codecov CLI to a secure, temporary directory and execute it from there. This removes the bypass check that allowed executing a local, potentially malicious 'codecov' binary from the untrusted workspace.

BUG=b/510376198
TAG=agy
CONV=f5118c4a-3577-4dc6-a4b6-1f2abb990935


khluu (approved):
This codecov thing is not actively being used at the moment, but it's good to have this anyway!

PR #390, opened 2026-06-24
[CI] Fetch latest main before computing merge-base to fix stale origin/main on workspaces

zxd1997066:
## Problem
On Buildkite agents with workspace reuse, `origin/main` can become stale if the local git state 
is not updated between builds. This causes `git merge-base origin/main HEAD` to compute an 
incorrect merge base, leading to:
- Incorrect file diff calculations for PR CI
- Pipeline step selection based on wrong change set

## Solution
Fetch latest main before computing merge-base, ensuring accurate diff and test selection

### DIFF
--- a/buildkite/pipeline_generator/utils_lib/git_utils.py
+++ b/buildkite/pipeline_generator/utils_lib/git_utils.py
@@ -9,6 +9,12 @@ def get_merge_base_commit() -> Optional[str]:
     merge_base = os.getenv("MERGE_BASE_COMMIT")
     if merge_base:
         return merge_base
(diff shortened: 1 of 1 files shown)


wendyliu235:
@khluu  Could you please review?

wendyliu235:
@zxd1997066  I would like to use this PR to fix it which only impact intel-CI. Please close your PR.
https://github.com/vllm-project/ci-infra/pull/392

zxd1997066:
Reopening, as it seems the pipeline generator v2 path still needs this change.

PR #392, opened 2026-06-24
Fix Intel CI merge-base diff calculation

wendyliu235:
## Summary

Fix Intel CI diff calculation to avoid using a stale `origin/main` from reused Buildkite workspaces.

The Intel bootstrap step previously computed `MERGE_BASE_COMMIT` directly from the local `origin/main` and recalculated diff with `git merge-base origin/main HEAD`. In PR builds where the checkout only fetched the PR head, `origin/main` could be stale, producing an oversized diff and unexpectedly setting `RUN_ALL=1`.

## Changes

- Fetch the PR base branch into `refs/remotes/origin/<base>` before computing `MERGE_BASE_COMMIT`.
- Compute `MERGE_BASE_COMMIT` once and export it for downstream pipeline generation.
- Use commit-to-commit diff: `MERGE_BASE_COMMIT..HEAD`.
- Remove `git add .` from diff calculation so workspace/index state does not affect changed-file detection.
- Use `LIST_FILE_DIFF=run_all` when all tests are already enabled, avoiding very large diff strings.

## Validation

- `bash -n buildkite/bootstrap-intel.sh`

### DIFF
--- a/buildkite/bootstrap-intel.sh
+++ b/buildkite/bootstrap-intel.sh
@@ -14,16 +14,16 @@ if [[ -z "${VLLM_CI_BRANCH:-}" ]]; then
     VLLM_CI_BRANCH="main"
 fi
 
+if [[ -z "${VLLM_CI_REPO:-}" ]]; then
+    VLLM_CI_REPO="vllm-project/ci-infra"
+fi
+
 if [[ -z "${AMD_MIRROR_HW:-}" ]]; then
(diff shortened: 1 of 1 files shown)


jikunshang (approved):

wendyliu235:
Vierfied PASS: https://buildkite.com/vllm/xpu-debug/builds/87#_

PR #395, opened 2026-06-25
[Bugfix] continue-on-failure exit status

AndreasKaratzas:
This fixes a regression in the Buildkite pipeline generator where failed commands were recorded in `__CI_OVERALL_STATUS`, but the generated script exited using `CI_OVERALL_STATUS`. Because those variables did not match, nightly steps could print failed test output while still returning exit code 0 to Buildkite. The change keeps the brace-based command wrapper from #381, so exports and other shell state still carry forward between commands. It only restores failure tracking to the same variable that the final `exit` uses. A regression test now runs a generated continue-on-failure script and verifies that later commands still run while the final process exits nonzero after a failed command.

### DIFF
--- a/buildkite/pipeline_generator/buildkite_step.py
+++ b/buildkite/pipeline_generator/buildkite_step.py
@@ -420,7 +420,7 @@ def _prepare_commands(
             if continue_on_failure:
                 # Note: We don't use a subshell here to preserve environment changes between commands
                 # (export, cd, etc).
-                commands.append(f"{{ {cmd}\n}} || __CI_OVERALL_STATUS=1")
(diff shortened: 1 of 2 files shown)


AndreasKaratzas:
Motivation was:
- https://buildkite.com/vllm/ci/builds/73953/canvas?sid=019ef838-3356-4df4-9b42-12035e696762&tab=output
- https://buildkite.com/vllm/ci/builds/73953/canvas?sid=019ef838-3364-47aa-b3c4-9f548f4e98af&tab=output

AndreasKaratzas:
Full CI run: https://buildkite.com/vllm/ci/builds/74379/canvas

PR #398, opened 2026-06-29
Use dedicated github app short lived token to checkout repo

theminghuang:
Previously we used github deploy key (ssh key) to access the private github repo.
However, deploy key is per repo and if we need to continue to the route, we need to mess with the vm ssh key everytime we set up a different repo. 
We happen to create a github app for managing vllm tpu related pr, let's reuse that instead. 
The way it works is:
1. we create a new script that can generate a short lived token based on github app private key
2. create a new buildkite hood that automatically invoke the script before checking out the repo
3. by the time we checkout the repo we will use the new github app token

Signed-off-by: Ming Huang <mhhua@google.com>

### DIFF
--- a/terraform/gcp_old/tpu-inference/modules/ci_cpu/main.tf
+++ b/terraform/gcp_old/tpu-inference/modules/ci_cpu/main.tf
@@ -48,7 +48,7 @@ resource "google_compute_instance" "buildkite-agent-instance" {
       #!/bin/bash
 
       apt-get update
-      apt-get install -y curl build-essential jq
+      apt-get install -y curl build-essential jq git python3 python3-pip
       sudo curl -L https://github.com/mikefarah/yq/releases/download/v4.53.2/yq_linux_amd64 -o /usr/bin/yq && sudo chmod +x /usr/bin/yq
 
       curl -o- https://get.docker.com/ | bash -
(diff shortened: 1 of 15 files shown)


theminghuang:

- #398 ◀

terraform/gcp_old/tpu-inference/modules/ci_cpu/main.tf
```diff
@@ -69,25 +69,59 @@ resource "google_compute_instance" "buildkite-agent-instance" {
...
+      # Setup In-Memory GitHub App Authentication
+      # ==========================================
+      pip3 install pyjwt requests cryptography --break-system-packages || pip3 install pyjwt requests cryptography
+
+      # 1. Create the Python script
+      cat <<'EOF' > /etc/buildkite-agent/get_github_token.py
+      import os, time, requests, jwt
+
+      APP_ID = "4156238"
+      INSTALLATION_ID = "142868369"
```
patrickji2014, line 82: The installation id might change if we reisntall the github app(not sure we will or not in the future though), anyway we can make it more flexible when it changes?

patrickji2014:
with this change, will all the machines has the env var GITHUB_TOKEN and we can use it in yml file? And iiuc, the token will expire, do we handle the refresh from infra side or as the GITHUB_TOKEN user, i need to refresh if it expires.

terraform/gcp_old/tpu-inference/modules/ci_cpu/main.tf
theminghuang: Yeah I checked your implementations and thought dynamically fetching the installation id logics is non trivial to maintain.
And the installation id is unique per organization and it seems unlikely for us to uninstall and reinstall on the org level for the foreseeable future (since we need to go through the vllm org every time). I'm inclined to hard code it as it for now. WDYT?

theminghuang:
> with this change, will all the machines has the env var GITHUB_TOKEN and we can use it in yml file? And iiuc, the token will expire, do we handle the refresh from infra side or as the GITHUB_TOKEN user, i need to refresh if it expires.

Synced offline but it should be fine as long as there's a dedicate step to interact with git (checkout, commit, create pr etc).

QiliangCui (approved):

QiliangCui (approved):

PR #401, opened 2026-07-03
Add tpu v7x-16 for buildkite agent

meiyeh123:
# Description

This changes add tpu v7x-16 for buildkite by using Terraform

Addition:
 - Use 'ci-bk' infix for multi-host TPU VM names and labels to avoid conflicts with existing v7x TPU resources
 - Include the hostname in Buildkite agent tags to better identify the agent.

FYI:
The multi-host does not support attaching _**read-write mode**_ disk ([ref](https://docs.cloud.google.com/tpu/docs/attach-durable-block-storage#access-modes)), and there isn't any way to scale up the vm boot disk size from even using the official `gcloud CLI`. 
Therefore, a better solution for accessing the model weights is using `runai_streamer` instead of downloading the weights to the local disk.

# Tests

[buildkite](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/22229#019f5945-0573-45a5-96ef-473ee9196bc8) (run the multihost test with the agent)

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.


CienetStingLin (approved):

terraform/gcp_old/tpu-inference/modules/ci_v7x_multi/main.tf
theminghuang, line 1: Looks like this is very similar to ci_v7x/main.tf file, do we see a way to update that startup script and reuse them?
Also it would be great that user don't have to remember if the instance they are creating are multi-host so that they don't accidentally set accelerator_type tpu7x-2 to a multi host module. https://screenshot.googleplex.com/C9Tjr6H4FAMcQyQ this seems to suggest anything larger than 8 cores is multi host

terraform/gcp_old/tpu-inference/modules/ci_v7x_multi/main.tf
```diff
@@ -0,0 +1,223 @@
...
+      mkdir -p /var/lib/buildkite-agent/.ssh
+
+      # Disable strict host checking for internal communication
+      cat <<EOF > /var/lib/buildkite-agent/.ssh/config
+      Host *
+        StrictHostKeyChecking no
+        UserKnownHostsFile /dev/null
+EOF
+
+      # Write the ephemeral private key generated by Terraform
```
theminghuang: The comment suggests the ssh key is the ephemeral but I think each time we run the cicd tests on multi hosts they need to communicate with each other. Would the key expire after certain time?

Lumosis (approved):

theminghuang (changes requested):
We also store cache (other than model weights that can be solved with runai_streamer) in the attached disk so I think attach disk is required

theminghuang:
> We also store cache (other than model weights that can be solved with runai_streamer) in the attached disk so I think attach disk is required

is READ_WRITE_SINGLE an option https://screenshot.googleplex.com/BQzwHVFX7LEacet? I think each host get it's own disk if I'm not mistaken and it can claim the sole ownership

meiyeh123:
> > We also store cache (other than model weights that can be solved with runai_streamer) in the attached disk so I think attach disk is required
> 
> is READ_WRITE_SINGLE an option https://screenshot.googleplex.com/BQzwHVFX7LEacet? I think each host get it's own disk if I'm not mistaken and it can claim the sole ownership

I don't think so. There's a note below:
https://screenshot.googleplex.com/8CBqCCHV66smRHf

And we have tested it. Multi-host actually does not work with single-writer mode

terraform/gcp_old/tpu-inference/modules/ci_v7x_multi/main.tf
meiyeh123: The key is considered _**ephemeral**_ because its lifecycle is tied strictly to the Terraform deployment. It is 
generated dynamically during apply, exists only within the state and the VM, and is discarded once the 
infrastructure is destroyed. This avoids the need for long-term secret management in GCP Secret Manager.

But we can tweak the comments if you think it's better.

theminghuang: Got it, yeah lets remove the ephemeral part since it's a bit misleading imo.

terraform/gcp_old/tpu-inference/modules/ci_v7x_multi/main.tf
```diff
@@ -0,0 +1,223 @@
...
+      # Forcefully intercept old CI jobs writing to /tmp and redirect them to the persistent disk
+      sudo rm -rf /tmp/tpu_jax_cache
+      sudo ln -s /mnt/disks/persist/tpu_jax_cache /tmp/tpu_jax_cache
+
+      # ==========================================
+      # 2. Automated Disk Garbage Collection (Cron)
+      # ==========================================
+      echo "Setting up daily cron job for JAX cache cleanup..."
+      echo -e '0 2 * * * root find /mnt/disks/persist/tpu_jax_cache -type f -mtime +30 -delete > /dev/null 2>&1\n30 2 * * * root find /mnt/disks/persist/tpu_jax_cache -type d -empty -delete > /dev/null 2>&1' | sudo tee /etc/cron.d/tpu_cache_cleanup
+      sudo chmod 0644 /etc/cron.d/tpu_cache_cleanup
```
theminghuang: Should we remove these and other setup related to /mnt/disks/persist as we don't really have a mount disk? And it also make it clear that for tests running on multi host should not use cache (or it will fully occupy the boot disk)

terraform/gcp_old/tpu-inference/modules/ci_v7x_multi/main.tf
meiyeh123: Done. Combine them and also extract it to `startup-script.sh.tftpl`

terraform/gcp_old/tpu-inference/modules/ci_v7x_multi/main.tf
meiyeh123: Done

theminghuang (approved):

CienetStingLin (approved):

PR #402, opened 2026-07-08
[CI][AMD] Add configurable native test execution without DinD

AndreasKaratzas:
## Summary

- Make `native_ci` an explicit runtime switch: `true` selects the native AMD Kubernetes pod, while `false` or omission preserves legacy DinD. ci-infra does not infer runtime mode from the device family or queue name.
- Complete the native pod contract with the per-commit `ci_base` image, explicit GPU/no-GPU quantities, 16 GiB `/dev/shm`, isolated workspace, `IPC_LOCK`, restricted security context, image pull policy, artifact/checksum parameters, job-local runtime settings, and Docker-hook suppression.
- Reject non-boolean `native_ci`, native multi-node jobs, and native `no_plugin` jobs during pipeline generation.
- Preserve the legacy runner for every AMD queue unless its configuration explicitly opts into native execution.
- Preserve upstream AMD mirror soft-fail behavior, per-step timeouts, and the latest upstream `main`.

This is paired with vllm-project/vllm#48387. The current vLLM configuration explicitly opts MI300 jobs into native execution while MI250, MI325, and MI355 remain on DinD. Future AMD queues can choose either mode without an infra code change.


khluu:
This would add to many things into `buildkite_step.py`. Can you make sure to utilize all other files like `constants.py` or `k8s_plugin.py`?

AndreasKaratzas:
CI runs with https://github.com/vllm-project/vllm/pull/48387:
- https://buildkite.com/vllm/ci/builds/77920/canvas
- https://buildkite.com/vllm/amd-ci/builds/10821/canvas

PR #406, opened 2026-07-09
Dump nvidia-smi topo -m for multi-GPU test steps

khluu:
## What

Emit `nvidia-smi topo -m` in the per-step setup for **NVIDIA steps that use more than one GPU** (tensor/pipeline parallel, i.e. `num_devices >= 2`, or multi-node `num_nodes >= 2`), right after the existing `nvidia-smi` GPU-info dump. It's gated to multi-GPU steps to keep single-GPU logs clean and wrapped in `|| true` so it never fails the step.

## Why

We recently chased a flaky "Fusion E2E TP2 Quick (H100)" failure that turned out to be **infra, not code**: the pod landed on a node whose NVLink/NVSwitch fabric was degraded, so every P2P-based communicator failed to init (`CUDA driver error: invalid device ordinal`) across three independent stacks (torch symmetric memory, vLLM custom all-reduce, FlashInfer workspace). The TP group fell back to PyNCCL, the AllReduce fusion pass disabled itself, and the tests failed their fusion-count assertions (`Could not find 2 ar_rms_fusion (found 0)`).

From the assertion errors alone this looked like a real fusion regression. The signal that would have made it obviously an infra flake — the GPU topology / P2P layout on that specific node — wasn't in the log. `nvidia-smi topo -m` captures exactly that, as the container sees it (which is what matters for P2P), so a degraded-fabric node is visible at a glance and flaky-by-scheduling failures are distinguishable from code regressions.

This only matters for jobs that actually use the fabric, hence the multi-GPU gate.

## Change

In `_get_setup_commands` ([buildkite_step.py](buildkite/pipeline_generator/buildkite_step.py)), for the `nvidia` setup profile:

```
echo '--- :nvidia: GPU Info'
(command nvidia-smi || true)
echo '--- :nvidia: GPU Topology'        # multi-GPU steps only
(command nvidia-smi topo -m || true)    # multi-GPU steps only
echo '--- :gear: CUDA Coredump Setup'
export CUDA_ENABLE_COREDUMP_ON_EXCEPTION=1 && ...
```

A small `_is_multi_gpu_step()` helper gates it on `num_devices >= 2` / `num_nodes >= 2`.

## Tests

Added three cases in `test_step.py`: topology present for a TP2 step, present for a multi-node step, and absent for a single-GPU step (base `nvidia-smi` still emitted). Full suite: 10 passed.

## Notes / possible follow-ups

- **Scope:** NVIDIA only, matching the class of failure above. AMD steps encode their GPU count in the device name (`mi300_4`) rather than `num_devices`, and would need `amd-smi topology` — left out to keep this focused; easy follow-up if wanted.
- This is purely diagnostic (log output); it does not change test selection, gating, or pass/fail behavior. A louder fail-fast/skip when P2P init fails on a queue where it's expected is a separate, bigger change.

### DIFF
--- a/buildkite/pipeline_generator/buildkite_step.py
+++ b/buildkite/pipeline_generator/buildkite_step.py
@@ -376,17 +376,42 @@ def _get_variables_to_inject() -> Dict[str, str]:
     }
 
 
+def _is_multi_gpu_step(step: Step) -> bool:
+    """Whether a step exercises more than one GPU.
+
+    Multi-GPU (tensor/pipeline parallel) and multi-node steps rely on the
+    cross-GPU fabric (NVLink/NVSwitch) and P2P being healthy on whichever node
+    they land on. Single-GPU steps don't, so the topology dump is only worth the
+    log noise for these.
+    """
+    if step.num_nodes and step.num_nodes >= 2:
+        return True
+    return bool(step.num_devices and step.num_devices >= 2)
+
+
 def _get_setup_commands(step: Step, setup_profile: SetupProfile) -> List[str]:
     if step.label.startswith(":docker:") or step.no_plugin or setup_profile == "none":
         return []
 
     if setup_profile == "nvidia":
-        return [
+        commands = [
             "echo '--- :nvidia: GPU Info'",
             "(command nvidia-smi || true)",
+        ]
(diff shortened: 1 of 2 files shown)


mgoin (approved):
Thanks!!

PR #418, opened 2026-07-14
[CI][AMD] Restore mirror failure policy

AndreasKaratzas:
- Restores `soft_fail` for AMD mirrors
- Removes the legacy Jinja renderer's implicit `amd_mi300` queue fallback


AndreasKaratzas:
Depends on:
- https://github.com/vllm-project/vllm/pull/48387

PR #429, opened 2026-07-22
[CI][AMD] Add opt-in Hugging Face Hub retries

AndreasKaratzas:
## Summary

Cached CI jobs should not contact Hugging Face Hub unless the cache is missing something. This change makes the AMD policy opt-in through `VLLM_CI_HF_HUB_MODE`. When the variable is unset or empty, both pipeline generators preserve the existing step environment and retry policy.

Using the Hub cache in online mode still performs metadata checks, which creates unnecessary traffic and exposes otherwise healthy jobs to rate limits and network timeouts. This PR supports an opt-in AMD rollout in both pipeline generators:

- Pass `VLLM_CI_HF_HUB_MODE` only to wrapped AMD test steps when it is explicitly enabled. Nightly and PyTorch-nightly builds promote an enabled mode to `online`.
- Retry exit status 75 once. The companion vLLM runner emits that status only for an offline cache miss or a transient Hub failure.
- Apply the same policy in the legacy MiniJinja template and the Python generator, including direct AMD-device steps and AMD mirrors.
- Leave the generated data unchanged when the gate is absent, including direct `no_plugin` commands, AMD infrastructure steps, the existing `AMD_RETRY` policy, and native source gating.

The error classification and request statistics live in vLLM; ci-infra only selects the mode and configures the Buildkite retry. Companion vLLM PR: https://github.com/vllm-project/vllm/pull/49401.

## Testing

- `pytest -q buildkite/tests` — 54 passed.
- `pytest -q buildkite/tests/test_amd_template_render.py` — 8 passed with MiniJinja 2.3.1.
- `ruff check buildkite/pipeline_generator/amd.py buildkite/pipeline_generator/buildkite_step.py buildkite/tests/pipeline_generator/test_amd.py buildkite/tests/test_amd_template_render.py` — passed.
- `git diff --check` — passed.

### DIFF
--- a/buildkite/bootstrap-amd.sh
+++ b/buildkite/bootstrap-amd.sh
@@ -179,6 +179,7 @@ upload_pipeline() {
             -D run_all="$RUN_ALL" \
             -D nightly="$NIGHTLY" \
             -D torch_nightly="$TORCH_NIGHTLY" \
+            -D hf_hub_mode="${VLLM_CI_HF_HUB_MODE:-}" \
             -D mirror_hw="$AMD_MIRROR_HW" \
             -D fail_fast="$FAIL_FAST" \
             -D vllm_use_precompiled="$VLLM_USE_PRECOMPILED" \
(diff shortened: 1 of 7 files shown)


AndreasKaratzas:
Closing as no traction.

PR #434, opened 2026-07-23
[CI][AMD] Fix ROCm base refresh timeout selection

AndreasKaratzas:
The AMD bootstrap collapses full-coverage diffs to `run_all` or `nightly`, which hid `docker/Dockerfile.rocm_base` from the refresh timeout selector and incorrectly produced 15 minutes. This preserves only the ROCm base Dockerfile marker alongside those bounded sentinels, retaining the existing 15-minute no-op and 540-minute rebuild policy without passing the full diff.


Hoiyeuhng:
**suggestion (non-blocking):** The fix itself looks good. One gap on the test side though: the new cases hand a pre-built `list_file_diff` straight to the Python generator, but the actual change is in `bootstrap-amd.sh` (appending the marker to the collapsed diff), and production still renders through the Jinja template. Neither of those is covered, so the marker preservation could regress while these tests stay green. Worth adding a test for that path?

AndreasKaratzas:
@Hoiyeuhng the Python pipeline does not share this bug.

PR #435, opened 2026-07-23
[CI][AMD] Add opt-in Hugging Face cache fallback

AndreasKaratzas:
- Gate the behavior at both pipeline and step level, emit the resolved runner mode explicitly, and use a fresh Buildkite retry so the selected presubmit job can rerun with Hugging Face clients online.
- Keep scheduled jobs online from their first attempt, exclude direct-command and multi-node jobs, and provide a strict emergency kill switch.
- Validate AMD mirror configuration with a typed schema and document the complete policy.


AndreasKaratzas:
Closing as no traction.

PR #438, opened 2026-07-27
tpu-inference/k8s: MultiKueue GKE infrastructure for TPU CI

theminghuang:
Terraform, Kueue manifest templates and a generator for running TPU CI on GKE. First of two stacked PRs: this one is the infrastructure a Buildkite step needs to land on a TPU node pool; [#467](https://github.com/vllm-project/ci-infra/pull/467) adds the launcher that turns a step into a workload. The pipeline that uses both is [tpu-inference#3377](https://github.com/vllm-project/tpu-inference/pull/3377).

## Shape

- One **manager** cluster (CPU only, `cloud-ullm-inference-ci-cd`, us-central1) running the Buildkite agent-stack-k8s controller on a single queue, Kueue, and the JobSet operator.
- One **worker** cluster per region (today `southamerica-west1-a` in `cloud-tpu-inference-test`) with TPU node pools that autoscale from the reservation, joined to the manager by MultiKueue over Connect Gateway.
- Per worker region: a models bucket and a compilation-cache bucket, exposed to workloads as PersistentVolumeClaims over GCS FUSE with fixed claim names, so a manifest never carries a bucket name.

## One input

`prod.auto.tfvars` declares the manager and a map of worker clusters, each with its TPU pools. The pool key is the profile name a pipeline names and, verbatim, the node pool, the ClusterQueue and the LocalQueue. Two consumers:

- `terraform apply` - clusters, node pools (reservation affinity, image streaming, TPU taint; a pool with `hosts > 1` is a multi-host slice pool with a placement policy carrying the topology, one node per host, scaled atomically), buckets and PVs, IAM (dedicated node SAs, Workload Identity for the workload SA and the Kueue controllers, cross-project Artifact Registry read), External Secrets, JobSet, Kueue.
- `scripts/generate_manifests.py` - cohorts and ResourceFlavors per accelerator, ClusterQueues and LocalQueues per pool (a worker's quota is its own capacity; the manager's is the sum over the workers that have the profile), MultiKueue fleet and AdmissionChecks, per-worker RBAC, the workload ServiceAccount, the cache volumes. Rendered into `generated/`, which is committed: a diff after an unrelated change means drift.

Only `google.com/tpu` is under Kueue quota; cpu and memory are ignored by admission (`quotaCheckStrategy: IgnoreUndeclared`) and left to the scheduler.

## The v6e worker

Two profiles on one 18-chip reservation: `v6e-1-1x1` (ten chips, two nodes kept warm) and `v6e-8-2x4` (one node). Multi-host slice pools (`hosts > 1`: placement policy with the topology, one node per host, atomic scaling, whole-slice validation) are supported but none is declared; a four-host `ct6e-standard-4t` 4x4 slice was run on this infrastructure during validation and is left out until multi-host steps join the suite.

## Verified

From tpu-inference: five consecutive full per-push suites (kube-dev builds 255-260, 16/16 steps), and separately a four-host JobSet slice built from zero, admitted as one workload, running JAX and vLLM TP=16 across hosts (build 272).


theminghuang:

- #438 ◀
    - #467

PR #448, opened 2026-08-03
[AMD][CI] Increase native shared-memory capacity

AndreasKaratzas:
- Increase the native AMD test pod's memory-backed `/dev/shm` size limit from 16 GiB to 32 GiB.
- Preserve the existing `emptyDir` mount and IPC locking configuration while giving large multi-GPU tests usable allocation headroom.

PR #46893 configured the DeepSeek KV-offload correctness test with a 16 GiB host cache, PR #37206 made the shared region populate its pages eagerly, and ci-infra PR #402 introduced the exact 16 GiB AMD pod limit. CUDA H100 CI can run that configuration because its memory-backed `/dev/shm` has no explicit 16 GiB cap, while AMD native CI currently caps the entire mount at exactly the requested allocation size. Build 11570 first exposes the job's missing AITER setting in the [KV-offload step](https://buildkite.com/vllm/amd-ci/builds/11570/list?sid=019fc12a-ba06-4b1d-bd34-32a2f7dcfbfd&tab=output); after enabling AITER, the exact local reproduction reaches this shared-memory limit. Raising the infrastructure limit preserves the original test coverage instead of reducing its cache size only for AMD.

### DIFF
--- a/buildkite/pipeline_generator/amd.py
+++ b/buildkite/pipeline_generator/amd.py
@@ -16,7 +16,7 @@ AMD_RESULTS_ROOT = "/home/buildkite-agent/huggingface/amd-ci-results"
 AMD_HF_HOME = "/home/buildkite-agent/huggingface"
 AMD_NATIVE_WORKSPACE = "/vllm-workspace"
 AMD_NATIVE_WORKSPACE_VOLUME = "vllm-workspace"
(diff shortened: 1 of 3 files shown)


AndreasKaratzas:
Closing as covered in https://github.com/vllm-project/ci-infra/pull/447

PR #459, opened 2026-08-07
[CI] Derive missing step keys from labels and reject duplicates

tahsintunan:
### Summary

50 steps across the CUDA and Intel pipelines declare no `key:`, which leaves them unaddressable. They cannot be re-run on their own, targeted by `depends_on`, or named in `VLLM_CI_ONLY_STEP_KEYS`, and the step selector skips them outright.

This PR derives a key from the step label when the YAML omits one, and fails pipeline generation if two steps end up sharing a key.


tahsintunan:
Closing in favor of #460

PR #462, opened 2026-08-11
refactor(tpu-inference/k8s): collapse duplicated Kueue queue templates

theminghuang:
queue_group_manager.yaml.tpl and queue_group_worker.yaml.tpl are identical
apart from the MultiKueue admissionChecksStrategy block. Every change to the
quota model - preemption policy, covered resources, flavor naming - has to be
made twice and kept in sync by hand, and there is nothing that catches a miss.

Replace both with queue_group.yaml.tpl plus an ${ADMISSION_CHECKS}
substitution the generator fills for the manager and leaves empty for workers,
and route both render paths through a shared render_queue_group() helper.

Verified as a pure refactor: regenerating from the committed prod.auto.tfvars
produces byte-identical output for every file under generated/, both the
modular parts and the consolidated bundles.

Also normalises hcl2 object keys. Keys written quoted in tfvars keep their
quotes through the parser and become Kubernetes object names, so a quoted
worker or pool key silently yields a ClusterQueue literally named
"v6e-1-1x1". The current tfvars uses bare identifiers so this is latent
rather than live, but it fails a long way from its cause. Confirmed both
spellings now generate identical manifests.

Adds scripts/requirements.txt; the generator imports hcl2 and nothing declared
it, so it fails with ModuleNotFoundError on a clean checkout.


theminghuang:

- #438
    - #462 ◀
        - #463
            - #464
                - #465

theminghuang:

PR #463, opened 2026-08-11
refactor(tpu-inference/k8s): single agent-stack controller on one queue

theminghuang:
With every TPU step going through the launcher, the Buildkite queue no longer
carries any information: the launcher submits the real workload and the TPU
profile is an argument to `launch`. A controller per profile was therefore
buying nothing, while making "add a TPU profile" mean "add a Helm release".

Collapse back to one controller on var.buildkite_queue, and drop
local.buildkite_queues along with the per-queue pod-spec-patch nodeSelector.
Placement - node selector, TPU resources, Kueue queue label - moves to the
submitted workload, where it belongs; the controller is now shape-agnostic and
adding a profile is a regenerated ConfigMap.

The agent pod is CPU-only and carries no Kueue queue label, so Kueue never
queues or evicts it. That is what makes the launcher worth having: the agent
acquires its Buildkite job in seconds instead of after admission and node pool
scale-up, and preemption evicts the workload without killing the build.

Also drops the custom controller image and the RBAC override. Both carried
buildkite/agent-stack-k8s#933, which shipped in v0.47.0, and the chart's
default Role at that version already grants delete on batch/jobs - the
override is now byte-identical to the default. Pin the chart and use the stock
image.

job-active-deadline-seconds and the workload deadline are derived from one
variable so they cannot drift apart.

### DIFF
--- a/terraform/gcp_old/tpu-inference/k8s/buildkite.tf
+++ b/terraform/gcp_old/tpu-inference/k8s/buildkite.tf
@@ -1,5 +1,24 @@
-# Buildkite Agent Stack Helm Release (Manager Cluster Only)
-# Note: Agent Stack controller is installed on manager only so worker clusters don't race to claim Buildkite jobs.
+# Buildkite Agent Stack (manager cluster only, so worker clusters never race
+# to claim Buildkite jobs).
+#
+# One controller, one queue. Every TPU step goes through the launcher, which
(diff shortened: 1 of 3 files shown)


theminghuang:

- #438
    - #462
        - #463 ◀
            - #464
                - #465

theminghuang:

PR #464, opened 2026-08-11
feat(tpu-inference/k8s): install JobSet and enable the Kueue integration

theminghuang:
A batch/v1 Job cannot span hosts, so neither multi-host TPU slices (one pod
per host, gang admitted as a unit) nor prefill/decode disaggregation
(independent server pods plus a benchmark pod) can be expressed today. JobSet
covers both, and is the object Kueue gang-admits.

Install the operator on the manager and on every worker, and add
jobset.x-k8s.io/jobset to integrations.frameworks in both Kueue configs.
MultiKueue mirrors the JobSet onto the selected worker, so the CRD, the
operator version and the enabled framework list all have to match on both
sides - hence a single jobset_version variable driving both installs.

Kueue only registers the integration if the JobSet CRDs already exist when its
controller starts, so both Kueue releases now depend on the JobSet release.

Two adjacent fixes in the same files:

  - kueue.tf pinned the chart version to a "0.19.0" literal in three places
    while var.kueue_version existed and was unused, so bumping Kueue meant
    editing the literals and the variable stayed silently wrong.
  - waitForPodsReady is inert in both configs: the block sets blockAdmission
    and timeout but never `enable: true`. Left off deliberately for now, but
    annotated so it does not read as active configuration.

(cherry picked from commit 700917f563430c56c01c3cbb90f33afb6286ed1b)

### DIFF
--- a/terraform/gcp_old/tpu-inference/k8s/jobset.tf
+++ b/terraform/gcp_old/tpu-inference/k8s/jobset.tf
@@ -1,0 +1,42 @@
+# JobSet operator.
+#
+# Required for any workload that spans more than one pod: multi-host TPU slices
+# (one pod per host, gang admitted) and prefill/decode disaggregation (separate
+# server pods plus a benchmark pod). A batch/v1 Job cannot express either.
+#
+# Installed on the manager and on every worker. MultiKueue mirrors the JobSet
+# object onto the selected worker, so the CRD and the operator must exist on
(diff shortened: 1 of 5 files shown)


theminghuang:

- #438
    - #462
        - #463
            - #464 ◀
                - #465

theminghuang:

PR #466, opened 2026-08-11
feat(tpu-inference/k8s): single-queue agent stack with JobSet support

theminghuang:
Brings the module to the shape TPU CI needs: one Buildkite queue served by one
controller, JobSet available for multi-pod workloads, and one Kueue queue
template.

Agent stack. A single controller on var.buildkite_queue. The Buildkite queue
carries no TPU information - the workload declares its own placement - so
adding a TPU profile does not mean adding a Helm release, and the controller
stays shape-agnostic. Its pod is CPU-only and carries no Kueue queue label, so
Kueue never queues or evicts it: the agent acquires its Buildkite job in
seconds rather than after TPU admission and node pool scale-up, and preemption
of a workload never takes the agent with it.

The chart is pinned to 0.47.0 and uses the stock controller image. 0.47.0 is
the first release containing the podless-Job cancellation fix
(buildkite/agent-stack-k8s#933), and its default Role already grants delete on
batch/jobs, so neither a custom image nor an RBAC override is needed.

JobSet. Installed on the manager and on every worker, with
jobset.x-k8s.io/jobset added to integrations.frameworks on both. A batch/v1 Job
cannot span hosts, so multi-host TPU slices and prefill/decode disaggregation
need JobSet, and MultiKueue mirrors the object across clusters - the CRD, the
operator version and the framework list all have to match on both sides, hence
one jobset_version driving both installs. Kueue only registers the integration
if the CRDs exist when its controller starts, so both Kueue releases depend on
the JobSet release.

The worker Kueue install now also triggers on its config content. A
null_resource local-exec re-runs only when a trigger changes, and the triggers
were version and cluster endpoint - so any edit to worker-config.yaml left
every worker on the old config while terraform reported no changes.

Queue templates. queue_group_manager and queue_group_worker were identical
apart from the MultiKueue admissionChecksStrategy block, so every change to the
quota model had to be made twice with nothing to catch a miss. One
queue_group.yaml.tpl with an ${ADMISSION_CHECKS} substitution, filled for the
manager and empty for workers. Verified as a pure refactor: regenerating from
prod.auto.tfvars produces byte-identical output for every file under
generated/.

Generator. hcl2 keeps the quotes on object keys written quoted in tfvars, and
those keys become Kubernetes object names, so a quoted key silently yields a
ClusterQueue named "v6e-1-1x1" - normalised, and both spellings now generate
identical manifests. Adds scripts/requirements.txt, since the generator imports
hcl2 and nothing declared it.


theminghuang:

- #438
    - #466 ◀
        - #467

PR #475, opened 2026-08-16
[CI][AMD] Upload failure diagnostics as Buildkite artifacts

AndreasKaratzas:
Configure AMD wrapper jobs to upload per-job failure diagnostics through the standard Buildkite post-command artifact phase.

- Add the exact `artifacts/amd-gpu-diagnostics/*/diagnostics.log` pattern to Python-generated AMD jobs.
- Configure the same path in the dedicated AMD pipeline template.
- Omit artifact collection for `no_plugin` jobs, which do not run the diagnostics wrapper.
- Set `BUILDKITE_ARTIFACT_UPLOAD_SKIP_SYMLINKS=true` so PR-controlled symlinks are never followed during artifact upload.
- Inject Kubernetes pod name, namespace, and node through Downward API fields for unambiguous diagnostics metadata.
- Retire the pre-#48646 commit-scoped ROCm image aliases; AMD jobs now use build-scoped `$BUILDKITE_BUILD_ID` tags unconditionally. Older vLLM branches must rebase before running current AMD CI.

### Companion rollout

This is phase 1 of a two-repository migration and must merge before [vllm-project/vllm#52264](https://github.com/vllm-project/vllm/pull/52264).

During the transition, current vLLM continues to write and upload `amd-gpu-diagnostics.log` inline, while this new `artifact_paths` pattern safely matches nothing. After #52264 lands, the runner writes `diagnostics.log`, records the injected Kubernetes metadata, and leaves artifact upload to Buildkite. This merge order avoids both artifact loss and duplicate uploads.

Validated together in:
- https://buildkite.com/vllm/amd-ci/builds/12112/list
- https://buildkite.com/vllm/ci/builds/84117/list

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,0 +1,2 @@
+[default.extend-words]
+hsa = "hsa"
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,2 +1,0 @@
-[default.extend-words]
-hsa = "hsa"
--- a/buildkite/bootstrap-amd.sh
+++ b/buildkite/bootstrap-amd.sh
@@ -163,12 +163,6 @@ upload_pipeline() {
     echo "AMD Mirror HW: $AMD_MIRROR_HW"
 
     FAIL_FAST=$(fail_fast)
-    ROCM_BUILD_SCOPED_IMAGES=0
-    if grep -q 'CI_BASE_IMAGE_TAG_BUILD_REF' \
-        .buildkite/scripts/ci-bake-rocm.sh 2>/dev/null; then
-        ROCM_BUILD_SCOPED_IMAGES=1
-    fi
-
     cd .buildkite
     (
         set -x
@@ -187,7 +181,6 @@ upload_pipeline() {
             -D vllm_ci_branch="$VLLM_CI_BRANCH" \
             -D rocm_base_refresh_skip="${ROCM_BASE_REFRESH_SKIP:-0}" \
             -D rocm_base_refresh_force="${ROCM_BASE_REFRESH_FORCE:-0}" \
-            -D rocm_build_scoped_images="$ROCM_BUILD_SCOPED_IMAGES" \
             | sed '/^[[:space:]]*$/d' \
             > pipeline.yaml
     )
--- a/buildkite/pipeline_generator/amd.py
+++ b/buildkite/pipeline_generator/amd.py
@@ -1,34 +1,14 @@
 import os
 from copy import deepcopy
 from dataclasses import dataclass
-from pathlib import Path
 from typing import Any, Dict, List, Mapping, Optional, TypedDict
 
 from constants import AgentQueue, DeviceType
 
 AMD_TEST_COMMAND = "bash .buildkite/scripts/hardware_ci/run-amd-test.sh"
 AMD_STABLE_CI_BASE_IMAGE = "rocm/vllm-dev:ci_base"
-AMD_LEGACY_NATIVE_BASE_IMAGE = "rocm/vllm-dev:ci_base-$BUILDKITE_COMMIT"
-AMD_BUILD_NATIVE_BASE_IMAGE = "rocm/vllm-dev:ci_base-build-$BUILDKITE_BUILD_ID"
-AMD_LEGACY_CI_IMAGE = "rocm/vllm-ci:$BUILDKITE_COMMIT"
-AMD_BUILD_CI_IMAGE = "rocm/vllm-ci:build-$BUILDKITE_BUILD_ID"
-
-
-def supports_build_scoped_images() -> bool:
-    producer = Path(".buildkite/scripts/ci-bake-rocm.sh")
-    return producer.is_file() and "CI_BASE_IMAGE_TAG_BUILD_REF" in producer.read_text()
-
-
-def get_amd_native_base_image() -> str:
-    return (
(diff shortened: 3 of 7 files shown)


AndreasKaratzas:
Test runs: 
- https://buildkite.com/vllm/amd-ci/builds/12112/list
- https://buildkite.com/vllm/ci/builds/84117/list

khluu:
Reviewed against vLLM `main`'s `.buildkite/scripts/hardware_ci/run-amd-test.sh` (the actual producer of these files). Existing generator tests pass (72/72).

### 1. The artifact glob never matches the file the wrapper writes (blocking)

`collect_rocm_failure_diagnostics` writes:

```
diagnostics_relative_path="${amd_diagnostics_dir}/${job_id}/amd-gpu-diagnostics.log"
```

so the file lands at `artifacts/amd-gpu-diagnostics/<job-id>/amd-gpu-diagnostics.log`, but both new globs expect `.../*/diagnostics.log`:

- `AMD_DIAGNOSTICS_ARTIFACT_GLOB` in `buildkite/pipeline_generator/amd.py`
- `artifact_paths` in `buildkite/test-template-amd.j2`

The post-command upload matches nothing and just logs a "no files matched" warning, so the feature is a silent no-op. Either use `*/amd-gpu-diagnostics.log` or widen to `artifacts/amd-gpu-diagnostics/**/*.log`.

### 2. The runner still uploads the same artifact inline

The comment change ("The runner uploads diagnostics immediately" → "Keep failure diagnostics out of shared HF/NFS storage") suggests the runner-side upload is gone, but on vLLM `main` `collect_rocm_failure_diagnostics` still ends with `buildkite-agent artifact upload "${upload_path}"`. Once the glob in (1) is fixed, every failing job uploads the same key twice. Worth landing the vLLM-side removal together with this, or dropping `artifact_paths` and keeping the inline upload.

### 3. Retiring commit-scoped aliases breaks branches that can't rebase

`bootstrap-amd.sh` fetches this template from ci-infra at pipeline-upload time and renders it against whatever vLLM branch is building. Branches like `releases/v0.11.2` have no `.buildkite/scripts/ci-bake-rocm.sh` at all, so their ROCm builds still push commit-scoped tags. After this change their native jobs get `image: rocm/vllm-dev:ci_base-build-$BUILDKITE_BUILD_ID` in `podSpecPatch` — an image nothing pushed — and every AMD step fails on ImagePullBackOff (the DinD fallback `rocm/vllm-ci:build-$BUILDKITE_BUILD_ID` is missing too). The PR description acknowledges this and says older branches must rebase, which release/backport branches generally can't do. If AMD CI is expected to keep working on release branches, the capability probe (or a `docker manifest inspect`-style fallback) should stay until those branches carry the build-scoped bake.

### 4. Pod name/namespace are injected but never recorded

`VLLM_CI_K8S_POD_NAME` and `VLLM_CI_K8S_NAMESPACE` are added to the Downward API env, but the wrapper only reads `VLLM_CI_K8S_NODE_NAME` (`amd_diagnostics_k8s_node_name`). The diagnostics log will still print only `k8s_node=...`, so the "unambiguous diagnostics metadata" goal needs the corresponding vLLM-side change.

### Checked and fine

- `BUILDKITE_ARTIFACT_UPLOAD_SKIP_SYMLINKS` is the correct env var for `artifact upload --upload-skip-symlinks`, and directory symlinks are already not followed during glob resolution by default.
- `artifact_paths: None` for `no_plugin` steps is stripped by `dict(exclude_none=True)` in `pipeline_generator.py`, so the serialized YAML stays clean — the added test covers it.
- The PR description still mentions a one-line typos entry in `pyproject.toml`; that commit was reverted in this branch, and `HSA` already passes on `main`, so the description is just stale.

AndreasKaratzas:
The missing context is that this PR is phase 1 of a coordinated two-repository migration with [[vllm-project/vllm#52264](https://github.com/vllm-project/vllm/pull/52264)](https://github.com/vllm-project/vllm/pull/52264), and this PR must merge first.

1. The `*/diagnostics.log` pattern is intentional. Current vLLM writes `amd-gpu-diagnostics.log` and uploads it inline, so during the transition the new automatic pattern safely matches nothing while the existing upload continues working. #52264 then changes the producer to `*/diagnostics.log`.

2. #52264 also removes the inline `buildkite-agent artifact upload`. In the final state there is only the standard Buildkite artifact phase, so there is no duplicate upload. Changing this PR to match the old filename would actually introduce duplication during the transition. The paired [[Buildkite run](https://buildkite.com/vllm/amd-ci/builds/12112/list)](https://buildkite.com/vllm/amd-ci/builds/12112/list) validated one automatic artifact per ordinary failed job.

3. The build-scoped-only image change is an intentional compatibility cutoff after vLLM #48646. A nightly does not make a pre-#48646 producer compatible: those checkouts do not publish the build-scoped native alias, and each run receives a different `$BUILDKITE_BUILD_ID`. The `v0.11.2` example is also already incompatible with the current template because it lacks other required ROCm build scripts. If current ci-infra is expected to support newer release branches that still predate #48646, that is a support-policy decision and we would need to retain the probe or backport the producer change.

4. The pod name and namespace are consumed by #52264, which records pod, namespace, and node in the diagnostics and clears the orchestration variables before running tests.

I’ve updated this PR’s description to document the companion PR, required merge order, transitional behavior, and paired validation runs. I also removed the stale HSA/typos bullet; no typos configuration is included in this PR.

PR #480, opened 2026-08-19
Source CI OTel helpers from the vLLM checkout

khluu:
## Summary

- remove the compressed/base64 OTel helper bundle from generated Buildkite commands
- resolve the helper directory once from the vLLM Git checkout and source it
- install no-op timing functions first, so missing or invalid helpers remain fail-open
- preserve trusted-build gating, command/test spans, OIDC upload, and original command exit behavior
- move implementation tests to the vLLM repository that now owns the helpers

Depends on vllm-project/vllm#52851 and must merge after it.

## Why

The helper archive accounted for roughly 9 KB of every traced `BUILDKITE_COMMAND`, making commands difficult to inspect. The setup fragment is now 384 characters, contains no embedded payload, and uses one checkout lookup instead of a list of runtime-specific paths.

Generated command wrappers are also smaller: they call the timing functions directly because setup always provides safe no-op fallbacks before attempting to source the real helper.

## Validation

- `PYTHONPATH=buildkite/pipeline_generator /Users/khluu/vllm-project/vllm/.venv/bin/python -m pytest -q buildkite/tests/pipeline_generator` — 86 passed
- `/Users/khluu/vllm-project/vllm/.venv/bin/pre-commit run --files buildkite/pipeline_generator/buildkite_step.py buildkite/tests/pipeline_generator/test_step.py` — passed
- generated setup regression asserts no `base64 --decode` and a command length below 500 characters
- cross-repository smoke test against the vLLM helper branch — passed and wrote one command span

## Rollout

1. Merge vllm-project/vllm#52851.
2. Merge this PR.
3. Run a trusted branch treatment build and verify command/test spans before normal main rollout.

All tracing setup remains soft-fail; missing files or helper errors leave the no-op functions in place and do not change the test command exit status.


khluu:
Verified end-to-end with the paired vLLM branch (vllm-project/vllm#52851):

- `buildkite/tests`: 87 passed (includes new regression test `test_otel_setup_stays_fail_open_without_git_checkout`, which fails without the `|| :` fix and passes with it).
- vLLM-side helper tests: `.buildkite/scripts/ci-otel/tests/test_ci_otel.py` — 30 passed.
- E2E in a `python:3.12-slim` container (git + pytest, dead OTLP endpoint) running the generated step script under `sh -e`:
  - generated step script is ~1.7 KB with no embedded base64 payload (was ~13 KB),
  - command spans (`ci.command`) and pytest spans (`pytest.test`) are recorded via the shim,
  - failing commands still propagate their exit code (exit 3 aborts the script as before),
  - no-`.git` / missing-helpers setups print `tracing setup skipped` and the job continues with exit 0.

Found and fixed one bug in the previous revision: the `git rev-parse` command substitution returned 128 outside a git checkout and, under `sh -e`, aborted the step before tests ran (commit da0fed8).

buildkite/pipeline_generator/buildkite_step.py
```diff
@@ -44,46 +39,42 @@
...
+    """Return True if cmd is safe to wrap with ci_otel_run.
+
+    ci_otel_run uses "$@" which works for external commands but not for shell
+    builtins that modify state (export, cd, etc.) or commands with shell
+    metacharacters (pipes, redirects, etc.).
+    """
+    first_word = cmd.split(None, 1)[0] if cmd.strip() else ""
+    if first_word in _SHELL_STATE_BUILTINS:
+        return False
+    return not any(char in _SHELL_METACHARS for char in cmd)
```
RichardoMrMu, line 80: [P1] Keep leading environment assignments in the shell

`_is_simple_command()` classifies commands such as `TP_SIZE=1 DP_SIZE=2 pytest ...` as simple, so the generated step becomes `ci_otel_run ... TP_SIZE=1 DP_SIZE=2 pytest ...`. After the paired helper shifts its two tracing arguments, `$@` tries to execute `TP_SIZE=1` as the program instead of applying it as a shell assignment, and the command exits 127 before pytest starts. This is reachable in current vLLM Buildkite jobs (for example the distributed, model-runner-v2, and rust-frontend suites all use this exact prefix, with additional single-assignment commands elsewhere). Please route commands with one or more leading POSIX assignment words through the explicit start/finish wrapper and add regressions for both single- and multi-assignment prefixes.

PR #486, opened 2026-08-25
tpu-inference: vllm-org CPU fleets, agent watchdog, name SSOT

theminghuang:
Groundwork for moving `tpu-inference` CI from the `tpu-commons` Buildkite org to the `vllm` org, plus fixes for an agent outage found along the way.

## Add parallel CPU fleets for the vllm org

New `ci_cpu` and `ci_cpu_64_core` fleets registered against the vllm org's agent token. These are **additive** — the tpu-commons fleets keep serving CI until traffic is cut over.

A new `purpose` variable switches a fleet onto self-describing names (`vllm-ci-cpu-64-core-vllm-us-central1-b-0`) shared by the VM, its boot disk, its static address, and the Buildkite agent, so a queue entry in the Buildkite UI maps straight onto a GCE instance. Empty `purpose` keeps the original names, so no existing fleet is renamed — and therefore none is recreated.

They run `n2d-standard-64` rather than `n2-standard-64`: n2 is persistently stocked out in us-central1, and n2d is the same 64 vCPU / 256 GB shape on AMD, so jobs need no retuning and `pd-balanced` still works.

## Keep the agent connected

Seven of eight CPU agents were found disconnected, with no alert. The packaged unit is `Restart=on-failure`, so it only recovers when the process exits non-zero — a server-side deregistration leaves the process alive, and systemd keeps reporting the unit active while the queue quietly drains.

Adds `modules/shared/keep-agent-connected.sh`, injected with `file()` into all four agent fleets. It sets `Restart=always` and installs a systemd timer that restarts the agent after three consecutive failures of the agent's own health endpoint. The endpoint stays healthy *during* a job, so a miss means wedged, not busy. The watchdog checks `systemctl is-enabled` first, so it no-ops on multi-host TPU workers that deliberately never join the queue.

This already caught a real failure during testing: a VM whose `buildkite-agent` install had flaked logged `health check failed 3 times, restarting agent`.

## Make the startup scripts reboot-safe

These scripts re-run on every boot, but were written as if they ran once.

- `sed s/xxx/<token>/` matched a pristine package placeholder that is gone by the second boot, so a rotated or re-pointed token never landed on a rebooted VM. Now rewrites the whole `token=` line.
- `tags=` and `HF_TOKEN=` were appended unconditionally, accumulating one duplicate per boot — a sampled VM had **eleven** of each. Now deleted before being written.
- `ci_cpu` has no `set -e`, so a flaked `apt-get install buildkite-agent` left a VM with no agent while the startup script reported success. Now asserts the package is installed.
- `ci_cpu_64_core` gains `allow_stopping_for_update` so the provider can resize a running instance.

## One local per module for resource names

The v6e/v7x name was spelled out four times per module — disk, VM, label, and again inside the startup script — and the copies had drifted. Multi-host v7x named the VM `tpu7x-16-ci-bk-0-…` while the agent inside it registered as `tpu7x-16-ci-0-…`.

All four now derive from one `node_names` local, and the v7x template no longer rebuilds the name at all — it receives it as `host_name`, so it cannot drift again.

That drift came from `ci_tmp`, added to dodge a collision with pre-existing v7x resources. Nothing named `tpu7x-16-ci-*` remains in the project, and the `accelerator_type` prefix already namespaces fleets against each other, so it is removed.

## Applied and verified

Already applied to both roots.

- `cloud-ullm-inference-ci-cd`: `2 added, 54 changed, 2 destroyed` — clean.
- `cloud-tpu-inference-test`: partial, see below.

The two adds/destroys are the `tpu7x-16` VMs, replaced by the `ci-bk` → `ci` rename. Both came back `READY`, and both agents reconnected with names matching their VMs:

```
GCE:       tpu7x-16-ci-0-cicd-us-central1-c   tpu7x-16-ci-1-cicd-us-central1-c
Buildkite: tpu7x-16-ci-0-cicd-us-central1-c   tpu7x-16-ci-1-cicd-us-central1-c
```

Reboot-idempotency was verified on a real stop/start: the VM came back with exactly one each of `tags=`, `token=`, `health-check-addr=`, and `HF_TOKEN=`, `Restart=always` in effect, watchdog timer active, and the health endpoint returning 200. Both rendered startup scripts were extracted from the plan and checked with `bash -n`.

Everything else is metadata-only and therefore **inert until each VM reboots or is recreated** — no running agent was disturbed.

## Known blockers, pre-existing and unrelated

Two items in `cloud-tpu-inference-test` could not be applied. Both predate this PR:

- **10 creates fail** — `ci_v6e_1` hits `Insufficient capacity` for v6e in us-east5-a, and `ci_v6e_8`'s `tpu_disk[8]` hits `Quota 'HDB_TOTAL_GB' exceeded` in southamerica-west1 (limit 98956 GB).
- **8 `ci_v6e_8` metadata updates** are stuck behind that disk: targeting them individually still pulls `disk[8]` in as a dependency. They will land once the quota is raised.

### DIFF
--- a/terraform/gcp_old/tpu-inference/cloud-ullm-inference-ci-cd/main.tf
+++ b/terraform/gcp_old/tpu-inference/cloud-ullm-inference-ci-cd/main.tf
@@ -3,6 +3,11 @@ data "google_secret_manager_secret_version" "buildkite_agent_token_ci_cluster" {
   version = "latest"
 }
 
+data "google_secret_manager_secret_version" "buildkite_agent_token_vllm" {
+  secret  = "projects/${var.secret_project_id}/secrets/vllm_buildkite_agent_token"
+  version = "latest"
+}
+
 data "google_secret_manager_secret_version" "buildkite_analytics_token_ci_cluster" {
   secret  = "projects/${var.secret_project_id}/secrets/tpu_commons_buildkite_analytics_token"
   version = "latest"
@@ -104,15 +109,16 @@ module "ci_v7x_16" {
   # disk_size defaults to 0, disable attached disk
 }
 
-module "ci_cpu_64_core" {
+# The whole tpu-commons 64-core fleet.
+module "ci_cpu_64_core_zone_c" {
   source = "../modules/ci_cpu_64_core"
   providers = {
-    google-beta = google-beta.us-central1-b
+    google-beta = google-beta.us-central1-c
   }
-
+  resource_suffix      = "-zone-c"
   project_id           = var.project_id
-  instance_count       = 4
-  machine_type         = "n2-standard-64"
+  instance_count       = 8
+  machine_type         = "n2d-standard-64"
   disk_size            = 250
   disk_type            = "pd-balanced"
   buildkite_queue_name = "cpu_64_core"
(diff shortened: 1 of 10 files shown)


theminghuang:

- #486 ◀

PR #488, opened 2026-08-27
[CI] Retire the tpu-commons half of CI monitoring

theminghuang:
## ⚠️ Merge inside the cutover window, Friday 2026-09-04

**After steps 30–31** have rolled the fleets onto the vllm agent token. Merged early, this stops monitoring an org that still has all the builds.

## What

Phase 4, step 32. Drops the `tpu-commons` entry from the two maps that drive monitoring:

| | Effect |
|---|---|
| `buildkite_token_secret_ids` | stops the `bk-metrics-tpu-commons` exporter; `tpu_commons/*` metrics stop advancing |
| `bq_puller_orgs` | stops polling the old org; rows already in `step_execution_logs` keep their `org_slug` and stay queryable |

Both orgs have been live in Stackdriver and BigQuery since #495 and #497, so this only retires the old half — nothing starts here. The Ops View dashboard can be repointed from `tpu_commons/*` to `vllm/*` ahead of the window, and rollback needs no monitoring action.

## Plan against live state

`0 to add, 2 to change, 2 to destroy` — both functions in-place, both `tpu-commons` IAM bindings destroyed. No source rebuild.

The metrics VM change is metadata-only, so the retired exporter keeps running until the box next boots. Harmless, and no in-window step is needed for it.

## Risk

Low mechanically, but correct **only after** the fleets have moved. Rollback is revert and re-apply.

### DIFF
--- a/terraform/gcp_old/tpu-inference/cloud-ullm-inference-ci-cd/main.tf
+++ b/terraform/gcp_old/tpu-inference/cloud-ullm-inference-ci-cd/main.tf
@@ -228,16 +228,12 @@ module "ci_monitoring" {
   project_id               = var.project_id
   bq_puller_pipeline_slugs = ["tpu-inference-ci", "vllm-torchtpu-ci"]
 
-  # Drop the tpu-commons entries once its agents are gone.
   buildkite_token_secret_ids = {
-    "tpu-commons" = "projects/${var.secret_project_id}/secrets/tpu_commons_buildkite_agent_token"
-    "vllm"        = "projects/${var.secret_project_id}/secrets/vllm_buildkite_agent_token"
+    "vllm" = "projects/${var.secret_project_id}/secrets/vllm_buildkite_agent_token"
   }
 
-  # vllm_buildkite_rest_api_token is the tpu-commons token despite its name.
   bq_puller_orgs = {
-    "tpu-commons" = "vllm_buildkite_rest_api_token"
-    "vllm"        = "vllm_org_buildkite_rest_api_token"
+    "vllm" = "vllm_org_buildkite_rest_api_token"
   }
 }
 


theminghuang:
Rebased onto `main` after #495 (both-org metrics coexistence) landed, and updated for two things that would each have broken monitoring silently.

**1. The metrics half is already done.** #495 made the exporter take a map of org slug → token secret and run one process per org, so `custom.googleapis.com/buildkite/vllm/*` has been populating since 2026-08-29 03:27Z. This PR no longer flips the agent token — it just drops the `tpu-commons` entry from the map, which is the actual cutover action.

That also means the Ops View dashboard can be repointed from `tpu_commons/*` to `vllm/*` **before** the window rather than during it, and rollback needs no monitoring action at all.

**2. The BigQuery puller needed a new REST token.** A Buildkite API token belongs to exactly one org, so flipping `org_slug` alone would have had the function call `/v2/organizations/vllm/...` with a token that can't see that org → 404 → `return "Error", 500` → the table quietly stops growing. Nothing would have caught it: the only alert policy in the project watches TPU disk utilisation, and the scheduler has no `retryConfig`.

The secret it was using, `vllm_buildkite_rest_api_token`, is unrelated to the vllm org despite the name — the "vllm" is the project. Created 2026-03-31, a personal token, 404s on the vllm org today. Now pointed at `vllm_org_buildkite_rest_api_token` (minted for this, verified 200 on vllm / 404 on tpu-commons), with the secret name hoisted into a module variable instead of being hardcoded.

**Plan on the current live state** — `1 to add, 2 to change, 2 to destroy`:

```
# google_cloudfunctions2_function.webhook_receiver          will be updated in-place
# google_compute_instance.monitoring_instance               will be updated in-place
# google_secret_manager_secret_iam_member.metrics_token_access["tpu-commons"]  will be destroyed
# google_secret_manager_secret_iam_member.secret_access     must be replaced
```

**Runbook note:** the instance change is metadata-only, and metadata updates don't reboot a VM. Without a stop/start the retired `tpu-commons` exporter keeps running until the next boot. That's benign — the `vllm` exporter is unaffected and metrics continue — so it needs no action inside the window; just don't expect the tpu-commons series to stop the moment this applies.

Still DO NOT MERGE until 2026-09-04.

theminghuang:
Rebased onto `main` after #497 merged; still held for 2026-09-04.

#497 made both the metrics exporter and the puller take a map of orgs, which subsumed the token hoist this PR used to carry. Both module files are untouched now, so this is a single-file change: drop one entry from each of the two maps. Rewritten as one commit rather than rebasing the old three.

PR #491, opened 2026-08-28
[CI] Gate native AMD jobs on ci_base-build image existing

sheralskumar:
- Add a CPU `verify-native-ci-base` step that checks `rocm/vllm-dev:ci_base-build-$BUILDKITE_BUILD_ID` exists on Docker Hub (`docker manifest inspect`) before native AMD jobs start.
- The template auto-adds that step to `depends_on` for every `dind: false` job. MI250 still uses DinD, so it only waits on `image-build-amd` and can run even if the native image check fails.

## Test plan 
Test on new `ci-infra` branch, using `VLLM_CI_BRANCH` env variable.
- [ ] Confirm native (`dind: false`) jobs wait on `verify-native-ci-base-amd` and skip if the tag is missing (no 15m pod timeout). (Example test case: https://github.com/vllm-project/vllm/commit/56e38f41ec7e5a3081febbaa713ef19607f72ff4 )
- [ ] Confirm MI250/DinD jobs still depend only on `image-build-amd`.
- [ ] Confirm the verify step finishes in seconds when the tag exists. (Example test case: https://github.com/vllm-project/vllm/commit/fd57c4b7afebc0b43d25ed7f5848fc35786463d0)


sheralskumar:
cc: @AndreasKaratzas

AndreasKaratzas:
Here is a test run with this branch:
- https://buildkite.com/vllm/amd-ci/builds/12453/list

AndreasKaratzas:
cc @Harry-Chen  @khluu for review

AndreasKaratzas:
I also snuck a modification for our CI since now we are consistently 98%+ so may as well remove the soft fail from our itnernal pipeline (even though it is not blocking)

PR #497, opened 2026-08-29
[CI] Poll both Buildkite orgs from the BigQuery puller

theminghuang:
The BigQuery puller read one org and one pipeline. This makes both a list, so `vllm-torchtpu-ci` and the `vllm` org land in `step_execution_logs` alongside the existing data instead of replacing it at cutover.

- `bq_puller_orgs` maps org slug => REST token secret. A Buildkite token is scoped to one org, so each brings its own; Terraform injects one secret env var per org.
- `bq_puller_pipeline_slugs` is a list. Every slug is polled in every org; one missing from an org is skipped rather than failing the run.
- Rows are stamped with a new nullable `org_slug` column. Existing rows keep `NULL`, and are all tpu-commons.
- `pipeline_slug`/`org_slug` renamed to `bq_puller_*`, since they drive only the BigQuery half of the module and not the metrics VM.

**Also fixes a data-loss bug.** Row IDs were `{build_id}_{step_name}`, but a step with `parallelism` emits several jobs sharing a name, so BigQuery deduped all but one away at insert. `vllm-torchtpu-ci` was worst hit: `Perf/Eval: General suite (TPU v7x-8)` runs `parallelism: 3` and kept 1 of 43 rows across 30 builds. Row IDs are now keyed on the job UUID. Rows already in the table are untouched.

Applied before merge. The BigQuery table updated in place — the column appends, `deletion_protection` untouched. Dry run across all four org/pipeline pairs: every row stamped, 0 row ID collisions, down from 50. A `vllm-torchtpu-ci` build has since landed in the table.

Rollback: revert and re-apply. The `org_slug` column is nullable and can stay.

### DIFF
--- a/terraform/gcp_old/tpu-inference/cloud-ullm-inference-ci-cd/main.tf
+++ b/terraform/gcp_old/tpu-inference/cloud-ullm-inference-ci-cd/main.tf
@@ -228,15 +228,22 @@ module "ci_monitoring" {
     google-beta = google-beta.us-central1-b
   }
 
-  project_id    = var.project_id
-  pipeline_slug = "tpu-inference-ci"
-  org_slug      = "tpu-commons"
+  project_id              = var.project_id
+  bq_puller_pipeline_slug = "tpu-inference-ci"
 
   # Both orgs are exported while the migration is in flight. Drop the
   # tpu-commons entry once its agents are gone.
   buildkite_token_secret_ids = {
     "tpu-commons" = "projects/${var.secret_project_id}/secrets/tpu_commons_buildkite_agent_token"
     "vllm"        = "projects/${var.secret_project_id}/secrets/vllm_buildkite_agent_token"
+  }
+
+  # A Buildkite API token belongs to a single org, so the puller needs one per
+  # org. vllm_buildkite_rest_api_token predates the migration and 404s on the
+  # vllm org despite its name.
+  bq_puller_orgs = {
+    "tpu-commons" = "vllm_buildkite_rest_api_token"
+    "vllm"        = "vllm_org_buildkite_rest_api_token"
   }
 }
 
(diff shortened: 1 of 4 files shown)


theminghuang:
Added `vllm-torchtpu-ci` to the puller, and applied.

`bq_puller_pipeline_slug` (string) becomes `bq_puller_pipeline_slugs` (list); every slug is polled in every org, so it is now a 2x2 matrix. A slug missing from an org 404s and is skipped, and the retry boundary moved from the org to the org/pipeline pair so one bad target cannot drop the others.

### This surfaced a pre-existing data-loss bug

Row IDs were `{build_id}_{step_name}`. A step with `parallelism` emits several jobs that **share a name**, so BigQuery deduped all but one of them away at insert. Measured over the last 30 builds per pipeline:

| pipeline | rows lost |
|---|---|
| `tpu-inference-ci` | 8 |
| `vllm-torchtpu-ci` | 42 — almost all from `Perf/Eval: General suite (TPU v7x-8)`, which runs `parallelism: 3` and kept 1 of 43 |

So the pipeline this PR adds is the one worst affected. Row IDs are now keyed on the job UUID, which is distinct per parallel job (`01a04bd3-368f-...`, `...-3692-482b-...`, `...-3692-401a-...` for `parallel_group_index` 0/1/2). Rows already in the table are untouched.

`timeout_seconds` 120 -> 240 for the extra sequential requests; still under the scheduler's 320s deadline.

### Verified

Applied (`1 added, 1 changed, 1 destroyed`, function in-place). Dry-ran the deployed `fetch_rows` over 24h across all four pairs: 2859 + 374 rows, every row stamped, **0 row_id collisions** (was 50). A forced scheduler run returned 200, and a `vllm-torchtpu-ci` build has since landed in the table with both `org_slug` and `pipeline_slug` populated — that pipeline had never appeared in `step_execution_logs` before.

Still no `vllm`-org rows: that org has produced one build ever on each pipeline. See the note below on its schedules.

PR #520, opened 2026-09-04
[CI] Accept generated AMD mirror keys in retry builds

Thangnguyenvn98:
## Problem

A cross-commit `/ci retry` forwards the source build's failed step keys via `VLLM_CI_ONLY_STEP_KEYS`. For an AMD mirror step, Buildkite emits the generated key `amd-<key>`, which the pipeline generator does not know (it only knows keys declared in `.buildkite/**/*.yaml`), so bootstrap dies with `Unknown CI step key(s)` and the retry produces nothing. Worse, the *next* retry uses the dead retry build as its source, sees only `bootstrap` failed, and retries `bootstrap` — the original failures are lost. (Incident: builds 87066 → 87079 → 87081.)

## Change

- `select_steps_and_dependencies` registers `amd-<key>` (with normalized AMD dependencies in a `dependencies_by_key` dict) for every step that has a `mirror.amd` block. A duplicate mirror key raises. The hard `ValueError("Unknown CI step key(s): …")` stays.
- In `convert_group_step_to_buildkite_step`, `include_step` gates the NVIDIA parent's block step and emission; the mirror condition becomes `only_step_keys is None or f"amd-{step_key}" in only_step_keys`.
- `_get_amd_mirror_effective_step` gives the copy `key=f"amd-{step.key or _generate_step_key(step.label)}"` so `_step_should_run` sees the mirror key in retry mode. Normal mode is unaffected (`amd-…` never matches the `image-build` prefix or `AMD_ALWAYS_RUN_STEP_KEYS`).
- `generate()` wraps the selection call in `try/except ValueError`, runs `buildkite-agent annotate <message> --style error` (`check=False`), and re-raises — the error shows on the build page above the job list instead of only in the bootstrap log.
- One sentence of README under `VLLM_CI_ONLY_STEP_KEYS`; one new test.

Diff: +83/−10 across 4 files.

## Tests

`python -m pytest buildkite/tests` → **104 passed** (103 existing unchanged + new `test_selected_steps_support_amd_mirror_keys`). Normal-mode pipeline output is byte-identical to `main`.

Manual-verification matrix (generator run from a vllm checkout): `VLLM_CI_ONLY_STEP_KEYS` unset → byte-identical pipeline; `["multi-modal-processor"]` → byte-identical; `["amd-multi-modal-processor"]` → previously `ValueError`, now emits `image-build-amd`, `ensure-ci-base-amd`, `refresh-rocm-base-amd`, `amd-multi-modal-processor` with no parent and no block step; `["bootstrap"]` → still `ValueError`, now also annotated on the build page.

Companion bot-side change (setup-failure comment, never retry `bootstrap`): separate PR against vllm-project/vllm, being opened concurrently; safe to merge in either order.

### DIFF
--- a/buildkite/tests/pipeline_generator/test_step.py
+++ b/buildkite/tests/pipeline_generator/test_step.py
@@ -119,6 +119,45 @@ def test_selected_steps_support_label_generated_keys():
     )
 
 
+def test_selected_steps_support_amd_mirror_keys(fake_global_config):
+    steps = [
+        Step(label="AMD image", key="image-build-amd", commands=["build"]),
+        Step(
+            label="Multimodal Processor",
+            group="Models - Multimodal",
+            key="multi-modal-processor",
+            commands=["test"],
+            mirror={
+                "amd": {
+                    "device": "mi355_1",
+                    "dind": False,
+                    "depends_on": ["image-build-amd"],
+                }
+            },
+        ),
+        Step(label="Other", key="other", commands=["other"]),
+    ]
+
+    selected, selected_keys = select_steps_and_dependencies(
+        steps, frozenset({"amd-multi-modal-processor"})
+    )
+    fake_global_config["only_step_keys"] = selected_keys
+    groups = buildkite_step.convert_group_step_to_buildkite_step(
+        group_steps(selected)
+    )
+    generated_keys = [job.key for group in groups for job in group.steps]
+
+    assert [step.key for step in selected] == [
+        "image-build-amd",
+        "multi-modal-processor",
+    ]
+    assert selected_keys == frozenset(
+        {"image-build-amd", "amd-multi-modal-processor"}
+    )
+    # The NVIDIA parent is neither emitted nor blocked; the mirror runs unblocked.
+    assert generated_keys == ["image-build-amd", "amd-multi-modal-processor"]
+
+
 def test_selected_steps_reject_duplicate_generated_key():
     steps = [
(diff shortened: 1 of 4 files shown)


buildkite/tests/pipeline_generator/test_step.py
```diff
@@ -119,6 +119,45 @@ def test_selected_steps_support_label_generated_keys():
     )
 
 
+def test_selected_steps_support_amd_mirror_keys(fake_global_config):
+    steps = [
+        Step(label="AMD image", key="image-build-amd", commands=["build"]),
+        Step(
+            label="Multimodal Processor",
+            group="Models - Multimodal",
+            key="multi-modal-processor",
```
tahsintunan: This only exercises the explicit-key path. Mirror keys are minted as `f"amd-{step.key or _generate_step_key(step.label)}"`, and the keyless branch is where the four real keyless mirrors land. Worth a second case with a keyless parent so `_generate_step_key` is pinned on both sides.

tahsintunan (commented):
LGTM. One non-blocking note inline. Happy to push the extra case if that's easier.

buildkite/tests/pipeline_generator/test_step.py
Thangnguyenvn98: I just made a new commit to re-adjust the test let me know if that good with you

tahsintunan: Yep, that works. Thanks for adding it!

PR #521, opened 2026-09-09
tpu-inference/k8s: the manager cluster

theminghuang:
First of eight PRs standing up a GKE + Kueue lane for tpu-inference CI alongside the bare-metal
Buildkite agent fleet. This one creates a regional **Autopilot** cluster in
`cloud-ullm-inference-ci-cd` that holds no accelerators: the Buildkite controller and the Kueue
control plane will live here, so neither competes for a chip with the tests it schedules, and a
worker cluster draining for maintenance cannot take the scheduler down. Nothing here is on any
pipeline's path.

Supersedes the manager half of #438, which stays open as the reference the split is cut from. The
PoC it replaces was torn down on 2026-09-08.

`terraform validate` passes. Not applied yet — that is the acceptance check for this PR.

### Why Autopilot

The steady state is ~2 vCPU of controllers plus one small launcher pod per in-flight step. The
Standard version of this cluster was three `e2-standard-4` nodes pinned up permanently, because a
regional pool needs a floor in every zone. When `queue: cpu` moves into the cluster (§8 of the
migration plan) the shape stays the same — a few hundred short, bursty, unprivileged jobs a day,
which is exactly what Autopilot is for and what a fixed pool sizes badly.

**Worker clusters stay Standard and must** — TPU pools need reservation affinity on a named
reservation, the `google.com/tpu` taint, COMPACT placement with an explicit topology, and
whole-slice atomic scaling. Those are the decisions Autopilot and node auto-provisioning take away.
Same reason NAP is not used here: the manager has one workload shape and no scheduling decision
worth delegating.

`enable_autopilot` is ForceNew, so it is settled before the first apply rather than after.

### The one Autopilot footgun, called out

Autopilot has no node pool to attach a service account to, which makes it easy to miss that its
nodes still run as one. The default is the project default compute SA — which on this project holds
`tpu.admin`, `storage.admin` and project-wide `secretmanager.secretAccessor`, because the agent VMs
share it. It is set here via `cluster_autoscaling.auto_provisioning_defaults.service_account`, pointing at a
service account this module always creates.

### Node service account

`roles/container.defaultNodeServiceAccount` plus `roles/artifactregistry.reader`, rather than the
older five-role recipe. The managed role is GKE's own definition of the node minimum and is strictly
narrower: it reaches the two metric *list* permissions directly instead of via
`roles/monitoring.viewer`, which would also grant read on every metric, dashboard and alert policy
in the project. `artifactregistry.reader` is separate because the managed role deliberately omits
it.

### What Autopilot costs us

Privileged containers, hostPath and host networking, so no `docker:dind` on the manager. This does
not constrain the plan: real work, CPU or TPU, is a Kueue workload on a Standard worker cluster, and
only controllers and launcher pods run here. Image builds are a separate decision, settled in §8.1 of the
migration plan: every viable builder on Autopilot is a form of delegation, because `buildkitd` needs
`CAP_SYS_ADMIN` and Kaniko is archived — so the builds go to **Cloud Build**, and no cluster here
ever needs a privileged container. (Mounting the host Docker socket is unavailable on
Standard too; GKE nodes have run containerd since 1.24.)

### Other deviations from #438

- `manager.tf` / `workers.tf` rather than one `clusters.tf`, so the file boundary matches the PR
  boundary and PR 2 is purely additive.
- State in the existing `tpu_commons_ci-infra_tf` bucket at prefix
  `terraform/cloud-ullm-inference-ci-cd-k8s-state`, matching the other three roots. All eight PRs
  share this prefix.
- Provider pinned to `google ~> 7.41`, lock file carrying `linux_amd64` as well as `darwin_arm64`.
- A nested `.gitignore` un-ignores `prod.auto.tfvars`; left ignored it needs `git add -f` on every
  edit.

### No `create_service_accounts` toggle

An earlier draft had `create_service_accounts` / `manager_node_service_account` so a deployment
could adopt a pre-existing account. That was for the test project, where a separate process wipes
Terraform-created resources; no resources remain there, so the flags encoded a hypothetical and cost
a three-way `coalesce` ending in the literal `"default"` plus a precondition guarding it. The module
now creates the account unconditionally. `terraform import` covers the adoption case if it returns.

### One thing to weigh in on

`artifactregistry.reader` is at **project** scope. It could be scoped to specific repositories, and
per our resource-IAM rule it should be. Left project-wide because it is not yet clear which repos
the manager pulls from, and the last attempt to narrow AR reader broke a build (64ca84f). Suggest
scoping once PRs 3 and 6 make the image list concrete.

### DIFF
--- a/terraform/gcp_old/tpu-inference/k8s/.gitignore
+++ b/terraform/gcp_old/tpu-inference/k8s/.gitignore
@@ -1,0 +1,4 @@
+# The repo root ignores *.tfvars as likely to hold secrets. This one holds
+# project ids, regions and CIDR blocks, and needs to be reviewed alongside the
+# code that reads it. Left ignored, every edit needs `git add -f` to commit.
+!prod.auto.tfvars
--- a/terraform/gcp_old/tpu-inference/k8s/.terraform.lock.hcl
+++ b/terraform/gcp_old/tpu-inference/k8s/.terraform.lock.hcl
@@ -1,0 +1,23 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/google" {
+  version     = "7.46.1"
+  constraints = "~> 7.41"
+  hashes = [
+    "h1:Hgv4BX9846ucZ6XcEql0DPfACoXqlcLat9SjUAn7YLI=",
+    "h1:oaPjmiPJz4rtO1CCcQ53YsY+Xr/vUEvmYRz10SGieuk=",
+    "zh:2e0e513915988d80582f8db7ff18c6242b8f4fad977d428feb11bcbd30d8ba95",
+    "zh:44ebe20d2a548c4318cc75497b7de0ded4bcb5867dbc9d7418f1941d5f6c5c2c",
+    "zh:48016a7ad372f27aa44b4f4611eb24fc634b05b0c25b84ca55a359f525f5a36f",
+    "zh:726fc1bbacf317bf946b53adc71248f890f78ede639c7d9e6fc52716668fd0ef",
+    "zh:9944d3128c69217d976844d936dd642d07140412d56e04bada95da189be37bdb",
+    "zh:a77548865bd320243641389763d3696b05c540d1cbb9868b249371626bab4132",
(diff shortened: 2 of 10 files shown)


theminghuang:

- #521 ◀

PR #527, opened 2026-09-10
tpu-ci k8s: stop indexing the node service account bindings by role

theminghuang:
There is one role, `container.defaultNodeServiceAccount`. A single-element `toset` crossed with the worker clusters gave `worker_node_role_bindings` — twelve lines of `flatten` to arrive at one binding per worker — and made both resources read as if the interesting question were which roles a node holds.

It isn't. The interesting question is that a node holds nothing beyond GKE's own minimum, because every pod on a node can reach the node's identity. That's now a comment on the resource instead of a structure implying a list that never grew.

Both addresses lose an index, so `moved` blocks keep Terraform from revoking and re-granting; the plan shows two moves and no other IAM change. They can be deleted after the next apply.

Stacked on #526.


theminghuang:
Folded into #526 — the stack was more ceremony than the change needed. The state re-index was done with `terraform state mv`, so the `moved` blocks are gone too.

PR #530, opened 2026-09-10
tpu-ci k8s: the Buildkite agent token, synced by GKE

theminghuang:
The Buildkite agent registration token as a Kubernetes Secret on the manager, plus read access to exactly that one Secret Manager secret. Fifth of eight PRs migrating tpu-inference CI to GKE + Kueue.

GKE's own `SecretSync`, not External Secrets: the reconciler runs in the control plane, so there is no Helm release, no Deployment, and nothing for the deploy script to install or wait on. It also drops the `tpu-ci-external-secrets@` service account ESO would have needed. Tested — the Secret appears within 20s and a new Secret Manager version lands in the same object within 270s. There is no watch, only a poll.

**Manager only**, and not because a worker does no Buildkite work — a TPU pod uploads its own artifacts and reports to Test Engine, since a pod shares no filesystem with the agent. It does that with `BUILDKITE_AGENT_ACCESS_TOKEN`, the per-job credential the agent gets back after registering, which the launcher forwards as an env var and which dies with the job. This secret is the *org registration* token; only the controller and the agent pods it creates present it, and both run on the manager. `secret_sync_config` is off on workers, so the CRD does not exist there either.

The sync has its own service account, not `tpu-workload` — reading the agent token and writing to the caches are different jobs, and only the second is ever mounted into a pod. Neither has a Google service account behind it. The grant is scoped to the one secret; the secret is a data source, not a resource, since it predates this fleet and its value is not Terraform's to hold.

Rotation reaching the Secret is not rotation reaching a process: the agent-stack-k8s controller holds its copy for the life of its process, so a token rotation still means restarting that Deployment. Noted for the PR 8 runbook.

Applied to `cloud-ullm-inference-ci-cd`: an `Opaque` Secret in `buildkite` with key `BUILDKITE_AGENT_TOKEN`, owned by the SecretSync, conditions `CreateSucceeded` / `UpdateNoValueChangeSucceeded`. Plan clean afterwards.


theminghuang:

- #530 ◀

PR #534, opened 2026-09-11
tpu-ci k8s: a README for the fleet

theminghuang:
Eight PRs built this and none left a page you could read first. Adds `terraform/gcp_old/tpu-inference/k8s/README.md`: what the two clusters are and why the work is split across them, what each path is, how to deploy a change, how a step asks for chips, and a runbook for adding a shape, rotating the agent token, rebuilding the launcher image, and tearing the fleet down.

The part that earns the page is "things that will surprise you" — the cluster-wide pod-creation blip while the Kueue controller rolls, the nodeAffinity Autopilot injects that makes a pod silently unschedulable on the worker, the short-workload log gap, and the fact that most of the agent-stack config is defaults we never wrote down.

Docs only. Every command in it was run against the live fleet, and the cluster names, queue names, bucket naming and webhook policies were read back from the cluster rather than from the code meant to produce them.

### DIFF
--- a/terraform/gcp_old/tpu-inference/k8s/README.md
+++ b/terraform/gcp_old/tpu-inference/k8s/README.md
@@ -1,0 +1,222 @@
+# tpu-inference CI on GKE
+
+TPU steps for the `tpu-inference` suite, run as Kubernetes workloads instead of
+on long-lived agent VMs. A Buildkite step on queue `kube` becomes a Kueue
+workload, which is admitted against fleet quota on a manager cluster and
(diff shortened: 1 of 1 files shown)


theminghuang:
Closing unmerged. A standalone README for the fleet would drift from the config it describes as this evolves, and a stale runbook is worse than none.

theminghuang:
Reopening — this README is in ci-infra (`terraform/gcp_old/tpu-inference/k8s/`), alongside the config it describes, not in the tpu-inference repo. Closed by mistake.

PR #535, opened 2026-09-11
ci(kube): let a manifest hold a role that wants no chips

theminghuang:
The 1P1D disagg benchmark is two vLLM servers holding v7x chips plus a proxy/benchmark pod holding none. The launcher refused it at submission — `resolve_shape` required every pod in a manifest to name the same shape.

Now only the roles holding chips decide the shape; a role that asks for no accelerator at all rides along, which is free because the queues put `google.com/tpu` alone under quota. Roles that do hold chips must still agree.

The docstring and README claimed a JobSet could put prefill and decode on two different shapes. That was never true and is not what this allows — both now describe the code.

First step of the PD disagg conversion; no capacity or deploy needed.

### DIFF
--- a/terraform/gcp_old/tpu-inference/k8s/README.md
+++ b/terraform/gcp_old/tpu-inference/k8s/README.md
@@ -76,10 +76,14 @@ states its hardware inside it:
 ```
 
 and passes nothing else. Not the shape, because a JobSet already says where each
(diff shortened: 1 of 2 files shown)


theminghuang:

- #535 ◀
    - #536
        - #537
            - #538
                - #539
                    - #540

PR #536, opened 2026-09-11
ci(kube): add a v7x worker cluster, and stop owning Cloud NAT

theminghuang:
Adds a second worker cluster — us-central1, because that is where the v7x reservation is and a node pool draws only on a reservation in its own zone. First time the fleet has been rendered with more than one worker; the names were already derived from (project, location), so the change is mostly the three things that broke.

- **Cloud NAT leaves this config.** A gateway covers every subnet range in its region and network, so it belongs to the region, not a cluster — and the manager already NATs us-central1, so declaring a second one here would be refused at apply. Routers/NATs are now created with the rest of the networking, named for what they serve. The four existing objects were `state rm`'d, not destroyed.
- **`vm.max_map_count` moves onto the TPU pools.** The existing lane sets it from a privileged init container in every manifest, which PodSecurity baseline on the workload namespace refuses. It is a node property.
- **The v7x pool** is 8 chips as two single-host 2x2x1 slices, `min_nodes = 0`. Not spare capacity — that reservation is fully consumed.

Plan is 15 to add, 0 to change, 0 to destroy. Not yet applied.


theminghuang:

- #535
    - #536 ◀
        - #537
            - #538
                - #539
                    - #540

PR #537, opened 2026-09-11
ci(kube): substitute a manifest once, not twice

theminghuang:
`render()` ran two `string.Template` passes over a manifest, so an escaped dollar had to survive both: `$$` collapses in pass 1 and is re-read as a placeholder in pass 2, meaning a single literal `$` had to be written `$$$$`. A manifest's containers are mostly shell, so this bites immediately.

One pass over the merged mappings. Required names still win, unknown names still raise, both hints unchanged — `$$` is now a literal `$`.

Prerequisite for writing the disagg JobSets.


theminghuang:

- #535
    - #536
        - #537 ◀
            - #538
                - #539
                    - #540

PR #538, opened 2026-09-11
ci(kube): let a manifest state how long its workload runs

theminghuang:
`tpu_test_max_seconds` becomes a default rather than a ceiling, and `tpu_total_max_seconds` becomes the ceiling — on both the deadline a manifest may ask for and the time the launcher will wait for admission.

A test has no opinion about its runtime and should get the fleet's number. A serving benchmark runs for as long as its client sweeps, which no shape can imply; the P/D disagg benchmark serves for ten hours and was being killed at three.


theminghuang:

- #535
    - #536
        - #537
            - #538 ◀
                - #539
                    - #540

PR #539, opened 2026-09-11
ci(kube): let the launcher read the Hugging Face token

theminghuang:
Resource-scoped `secretAccessor` on `bm-agent-hf-token` in `cloud-tpu-inference-test` for the launcher's service account, plus the two variables naming it.

A gated model cannot be fetched without it, and the fleet's model cache is shared, so the first step to want one pays the download for every later step. The bare-metal agents' secret rather than a new one: a gate is per account and both lanes pull the same weights.


theminghuang:

- #535
    - #536
        - #537
            - #538
                - #539 ◀
                    - #540

PR #540, opened 2026-09-11
ci(kube): move the manager cluster off Autopilot

theminghuang:
Rebuilds `tpu-ci-manager` as a Standard cluster with its own node pool, so the cluster that builds workload podspecs is not one that rewrites them.

Autopilot fills in a `cloud.google.com/extended-duration-pods` nodeAffinity on any pod that lacks one; MultiKueue copies the podspec to a worker where no node carries that label, and the pod is silently unschedulable.

ForceNew, so the cluster is recreated. Everything on it is generated YAML that `deploy_manifests.py` reapplies.


theminghuang:

- #535
    - #536
        - #537
            - #538
                - #539
                    - #540 ◀

PR #541, opened 2026-09-11
ci(kube): say what --env may forward

theminghuang:
`forward_env` never had an allowlist, but its docstring read as though it did. The kube P/D disagg lane needs the workload pod to upload its own results, which means forwarding the agent's job credential by name — so the rule is that a forward is per-name and visible in the step, not that it cannot happen.

Docstring only; no behaviour change.


theminghuang:

- #535
    - #536
        - #537
            - #538
                - #539
                    - #540
                        - #541 ◀
                            - #542
                                - #543
                                    - #545
                                        - #546
                                            - #547

PR #542, opened 2026-09-11
ci(kube): pin a workload's image to one digest

theminghuang:
A tag can be republished between a run's pulls — a pod per role, plus every restart — which puts two builds in one benchmark and invalidates a compile cache keyed on the image. The launcher now resolves a tag to its digest at submission, so every pod in a workload gets the same bytes.

This is what the vllm-torchtpu disagg wrapper script did before calling the launcher; nothing about it was that pipeline's.

Also grants `artifactregistry.reader` to the launcher's KSA — the existing grants are on the node service accounts, which is what pulls an image, not what queries one.


theminghuang:

- #535
    - #536
        - #537
            - #538
                - #539
                    - #540
                        - #541
                            - #542 ◀
                                - #543
                                    - #545
                                        - #546
                                            - #547

PR #543, opened 2026-09-11
ci(kube): let the launcher supply a fleet secret to --env

theminghuang:
The HF token grant from #539 had nothing reading it — the value reached a workload only because the vllm-torchtpu disagg lane fetched it with gcloud in a wrapper script. That put which-secret-in-which-project in a pipeline, and made every lane wanting a gated model repeat it.

The launcher now resolves a `--env NAME` it does not have set against `env_secrets` in the profile registry. Still opt-in per name; the ConfigMap carries the secret's name, not its value.

---

Also carries #544 ("Auto-provision the manager's nodes instead of pinning a pool"), which was squash-merged into this branch rather than into `main`. Merging this PR lands both.


theminghuang:

- #535
    - #536
        - #537
            - #538
                - #539
                    - #540
                        - #541
                            - #542
                                - #543 ◀
                                    - #545
                                        - #546
                                            - #547

PR #545, opened 2026-09-11
Size TPU node pools per zone, so a multi-host slice is valid

theminghuang:
The `tpu7x-standard-4t-2x2x2` pool would not create: *"Maximum node count 0 is not a valid size of TPU pod slice with topology 2x2x2"*. GKE validates a multi-host slice against the per-zone `max_node_count`, which is zero when `total_max_node_count` is used instead.

Each pool is pinned to one zone by `node_locations`, so the two fields carry the same number and nothing else changes. `location_policy = "ANY"` goes with it, since a balanced spread cannot assemble a slice.


theminghuang:

- #535
    - #536
        - #537
            - #538
                - #539
                    - #540
                        - #541
                            - #542
                                - #543
                                    - #545 ◀
                                        - #546
                                            - #547

PR #546, opened 2026-09-11
Drop the affinity that kept Autopilot out of a podspec

theminghuang:
The launcher wrote a nodeAffinity into every pod it submitted, restating the manifest's own nodeSelector, so that Autopilot would not add `cloud.google.com/extended-duration-pods` — a label no Standard worker node carries, on a podspec MultiKueue copies across verbatim.

The manager is Standard now and every worker always was, so nothing in the fleet mutates a podspec. The affinity constrained nothing the nodeSelector and the TPU taint did not already. The README keeps the failure mode, restated as a constraint on what the fleet may be built from.


theminghuang:

- #535
    - #536
        - #537
            - #538
                - #539
                    - #540
                        - #541
                            - #542
                                - #543
                                    - #545
                                        - #546 ◀
                                            - #547

PR #547, opened 2026-09-11
Place a tpu7x slice with a workload policy, not a compact one

theminghuang:
The `tpu7x-standard-4t-2x2x2` pool created and then went straight to ERROR: *"Creation of a managed instance group with tpu7x-standard-4t machine type with placement policy is not supported. Use workload policy instead."*

tpu7x takes a named `HIGH_THROUGHPUT` workload policy carrying the accelerator topology, in place of the compact placement policy GKE creates for earlier generations from `--tpu-topology`. Confirmed against the live cluster first: a probe pool came up RUNNING and the API stored `policyName` and `tpuTopology` with no `type`, which is why `type` is empty here — set, it would diff against what GKE reports and replace the pool on every apply.

The branch is on the machine family, so a future multi-host v6e pool keeps the compact path.

### DIFF
--- a/terraform/gcp_old/tpu-inference/k8s/locals.tf
+++ b/terraform/gcp_old/tpu-inference/k8s/locals.tf
@@ -70,6 +70,10 @@ locals {
             is_multi_host = dims[0] * dims[1] * dims[2] > parseint(
               trimsuffix(reverse(split("-", pool.machine_type))[0], "t"), 10
             )
+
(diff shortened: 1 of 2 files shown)


theminghuang:

- #535
    - #536
        - #537
            - #538
                - #539
                    - #540
                        - #541
                            - #542
                                - #543
                                    - #545
                                        - #546
                                            - #547 ◀

PR #550, opened 2026-09-11
ci(kube): pin a workload's image to one digest

theminghuang:
Re-lands #542, whose squash merge went into `launcher-env-forwarding-doc` rather than `main` and so never reached the trunk. Same commit, cherry-picked onto current main; no content change.

The launcher resolves a step's image tag to a digest at submission, so every pod in a run uses the bytes that existed when the run started, even if the tag moves under it.

This is already what the deployed launcher does — the manifests were applied from a branch that had it — so this closes a gap between the cluster and the trunk.


theminghuang:

- #550 ◀

PR #551, opened 2026-09-11
Drop the affinity that kept Autopilot out of a podspec

theminghuang:
Re-lands #546, which was **closed rather than merged**: deleting `worker-tpu-multihost-autoscaling` on #545's merge closed the PR stacked on it instead of retargeting it. Same commit, cherry-picked onto current main; no content change.

Touches README.md and launch.py only, so it does not overlap #552 and the two can merge in either order.


theminghuang:

- #551 ◀

PR #552, opened 2026-09-11
Place a tpu7x slice with a workload policy, not a compact one

theminghuang:
Re-lands #547, whose squash merge went into `launcher-drop-autopilot-affinity` rather than `main` and so never reached the trunk. Same commit, cherry-picked onto current main; no content change.

Touches locals.tf and workers.tf only, so it does not overlap #551 and the two can merge in either order.


theminghuang:

- #552 ◀

PR #563, opened 2026-09-12
tpu-inference/k8s: give a cold image pull longer than thirty seconds

theminghuang:
Raises the agent-stack-k8s controller's `image-pull-backoff-grace-period` from
the chart default of 30s to 5m, in both the template and its generated output.

The default assumes a warm node and a small image. A fleet node is usually
brand new and the image is a CI build from the commit under test, so the pull
is always cold. Steps were failing with exit -1, no agent and an empty log:
the controller had given up before a pod ever ran, which presents as a
Buildkite fault rather than a slow pull.

Already applied and deployed to the manager cluster.


theminghuang:
Opened against a stale remote ref; superseded by #564.

PR #600, opened 2026-09-17
Skip AMD steps on TORCH_NIGHTLY full CI runs

divakar-amd:
## Summary

`TORCH_NIGHTLY=1` runs (e.g. https://buildkite.com/vllm/ci/builds/89623) validate vLLM against a CUDA-only torch-nightly build, and rely on the existing full-run semantics to force every step to run — including the AMD image-build lane and every AMD-mirrored test step. ROCm has its own pinned torch and gains no signal from validating against a CUDA torch-nightly build, so this just burns AMD CI capacity for no reason.

This PR adds `drop_amd_steps()`, applied once in `PipelineGenerator.generate()` right after steps are collected from `job_dirs` (before grouping/selection), when `TORCH_NIGHTLY == "1"`:
- Drops `amd_cpu` build steps (`ensure-ci-base-amd`, `image-build-amd`, `refresh-rocm-base-amd`).
- Drops native AMD GPU test steps (any `device` matching `is_amd_gpu_device`).
- Strips the `"amd"` key out of any step's `mirror` dict, so no `AMD: <label>` companion step / `Hardware-AMD Tests` group is generated for that step — while leaving the primary (non-AMD) step and any other mirror keys untouched.

Everything else continues to run via the existing `TORCH_NIGHTLY` full-run behavior; only the AMD lane is excluded.

## Test plan

- Added unit tests in `buildkite/tests/pipeline_generator/test_step.py` covering: `amd_cpu`/native-AMD-GPU steps being dropped, `mirror.amd` being stripped while the primary step and other mirror keys are preserved, and an end-to-end check that a `TORCH_NIGHTLY=1` pipeline generation produces no `Hardware-AMD Tests` group and no `amd-*` step keys.
- `PYTHONPATH=. python -m pytest ../tests/pipeline_generator/ -q` from `buildkite/pipeline_generator/` — 112 passed (108 pre-existing + 4 new).


khluu (commented):
The current approach works, but this could be simpler: compute `include_amd = global_config["torch_nightly"] != "1"` in `convert_group_step_to_buildkite_step()`, skip AMD devices in its loop, and gate AMD mirror creation with the same flag. Keep `is_amd_device()` and remove `drop_amd_steps()` and the mirror-copying pass. I prototyped this against current vLLM definitions: identical full torch-nightly output, with normal full runs unchanged.

Two follow-ups:
- Add a generation-level test for nightly on/off; the helper tests don't exercise the flag wiring. Preserve an explicit error for requested AMD keys under torch-nightly rather than silently dropping them.
- Make the annotation best-effort or use logging: `check=False` still raises `FileNotFoundError` when `buildkite-agent` is absent (reproduced).

Validation: 111 tests passed; the PR removed 157 AMD entries, preserved all retained groups, and left no dangling dependencies. No blocking issue found in the scheduled-run path.

divakar-amd:
Thanks for the review @khluu. Updated the PR. Looking for your review again.

PR #617, opened 2026-09-22
buildkite-agent hooks: exit 255 on provider/infra failures, and retry 255

Thangnguyenvn98:
## Problem

[vllm/ci build 90241](https://buildkite.com/vllm/ci/builds/90241) flaked: a hook on an H200 MIG agent hit a transient provider failure (AWS CLI / network blip) and exited 1. The generated steps only auto-retry on `exit_status: -1` (agent lost), so the hook's exit 1 was not retried and one blip failed the whole build.

## Fix

Two parts, because Buildkite preserves a hook's exit code as the job's exit status (see agent `internal/job/executor.go` — a hook's typed `ExitError` code is returned verbatim; `-1` is reserved for agent-lost):

1. **Hook templates** (`scripts/buildkite-agent/`) adopt an explicit exit-code convention:
   - **exit 1 — genuine misconfiguration** (wrong `spawn` count, stale CDI spec, invalid git config state). Fails immediately, as before.
   - **exit 255 — provider/infrastructure failure** (AWS CLI / ECR login, registry blip, secret-store fetch, `docker pull`, `nvidia-smi` hiccup).
2. **Pipeline generator** (`buildkite_step.py`): `ensure_infra_failure_retry` (renamed from `ensure_exit_status_negative_one_retry`) now prepends a one-time `exit_status: 255` automatic retry alongside the existing `-1` rule on every generated step, so a provider-side hook failure re-runs instead of failing the build.

Guarded as 255 (provider/infra):

- `environment`: both ECR logins (AWS CLI + registry), `docker pull` of the job image, `nvidia-smi -L` failure (driver hiccup)
- `pre-checkout`: `buildkite-agent secret get GITHUB_READONLY_PAT` (secret store unreachable)

Left as 1 (our fault, retrying won't help): bad/missing agent number, agent number exceeding the MIG devices the host actually has, CDI device absent from the spec, invalid `GIT_CONFIG_COUNT`.

## Tests

- `bash scripts/buildkite-agent/tests/pre-checkout.sh` — extended with a `secret-store-down` case asserting exit 255; all 12 cases pass.
- `environment` hook smoke-tested locally with stubbed `aws`/`docker`/`nvidia-smi`: AWS CLI failure, registry login failure, and `nvidia-smi` failure all exit 255; `nvidia-smi` succeeding with no MIG slices (misconfiguration) still exits 1.
- `pytest buildkite/tests` — 130 passed, including updated retry-policy expectations in `test_step.py` / `test_l4_k8s.py`.
- `pre-commit run` clean.

## Deployment

Hook templates are not auto-deployed: after merge, re-install `environment` and `pre-checkout` on the affected agents with the Quick install commands in `scripts/buildkite-agent/README.md`. Hooks are re-read at every job start, so no agent restart is needed. The pipeline-generator half takes effect on the next pipeline upload.


buildkite/pipeline_generator/buildkite_step.py
```diff
@@ -603,23 +607,23 @@ def ensure_exit_status_negative_one_retry(
...
-        ):
-            return retry_policy
     else:
         raise ValueError("retry.automatic must be a boolean, mapping, or list.")
 
-    retry_policy["automatic"] = [
-        dict(EXIT_STATUS_NEGATIVE_ONE_RETRY),
-        *automatic_conditions,
+    existing_statuses = {
+        condition.get("exit_status")
```
khluu: [P2] This set comprehension crashes on valid retry rules with an exit-code array, e.g. `exit_status: [1, 2]`, because lists are unhashable. I reproduced `TypeError: unhashable type: 'list'` on this head; the same input succeeds on the base revision. Since this runs during step generation, one such rule aborts the entire pipeline upload. Please normalize array values before deduplication and add a regression test. Buildkite documents array support here: https://buildkite.com/docs/pipelines/configure/retry#automatic-retry-attributes

Thangnguyenvn98: Fixed in 4e8b741: array values are normalized into the status set before dedup, so `exit_status: [1, 2]` no longer raises `TypeError: unhashable type: 'list'`. Added two regression tests — one for the crash (`test_infra_retry_tolerates_exit_status_arrays`) and one for dedup when an array already contains 255 (`test_infra_retry_dedupes_statuses_inside_arrays`).

PR #619, opened 2026-09-22
kernrec: honor the symbol map's empty-with-reason contract in the consumers

khluu:
Follow-up to #611, which was merged before this commit landed on its branch.

The map producer (vllm-project/vllm#58097) fails closed: if any object is unreadable after a retry, the map goes out with an empty object list and a `reason`. `kernel_table.py query` and `join_symbol_map.py` now say so explicitly and fall back to the static rule for every file. They would also treat files touched by an object carrying `error` as undecidable, should a producer ever publish a partial map.

Tooling only; nothing in CI reads these yet.


khluu:
Superseded: the same commit is included in the collect-step PR (kernrec-collect branch), which also needs it.

PR #620, opened 2026-09-22
[CI] Collect the kernel coverage table at the end of recording builds, publish to S3

khluu:
## What

A recording build (`VLLM_CI_KERNREC=1`, i.e. the nightly and daily full runs) now ends with a **`kernrec-collect`** step that folds every job's kernel recordings into the per-step kernel table, pairs it with the build's kernel symbol map, and publishes both.

- **No Buildkite API needed.** `ci_setup.sh` writes a `kernrec.json` sidecar on exit (step key, shard, exit status). The collect step downloads this build's own `.fnrec/**/*` artifacts and files each job under its step. Jobs killed before their trap ran have no sidecar and count as not passed.
- **`kernel_table.py build --fnrec`** reads that layout (the offline step layout is unchanged) and runs on the agents' Python 3.9.
- **`collect.sh`** uploads the table and the map as artifacts of the collect job and, when the agent has AWS identity, to `s3://vllm-ci-selector/<pipeline>/<commit>/` plus `latest.json`. Publishing problems are reported, never fatal.
- **Generator:** appends the step when recording is on. It depends on every command step that will run (blocked steps excluded, so it can never wait on a step that will not start), `allow_dependency_failure` so failed jobs still contribute evidence, `soft_fail` so publishing never turns a build red. Small CPU postmerge queue on main, premerge otherwise.
- **Terraform:** bucket `vllm-ci-selector` with public read (bootstrap will fetch over HTTPS with no credentials; contents are step keys and kernel names), a 365-day lifecycle, and a write policy attached to the postmerge agent stacks that run the collect step. **Needs `terraform apply`**; until then the step publishes artifacts only and says so.

Includes the consumer-side empty-map fallback from #619 (supersedes it).

## Verified

- 140 tests pass, including new ones for the sidecar layout, failed-job rows, the block-step exclusion, and the queue choice.
- `test_checkout_perms.sh` (Docker) still passes with the reworked exit trap: normal exit, SIGKILLed step, and the pre-fix control.
- Rendered a pipeline with recording on: the collect group is last, depends on every command step, `allow_dependency_failure: true`, `soft_fail: true`, `small_cpu_queue_postmerge` on main.
- Shellcheck clean on `collect.sh` and `ci_setup.sh`.

A CI run of the collect step on a fork branch follows in a comment.


khluu:
Fork-branch CI run of the collect step, as promised: [vllm/ci build 90406](https://buildkite.com/vllm/ci/builds/90406) (`khluu:kernrec-collect-test`, vLLM `91d7324c`, `VLLM_CI_KERNREC=1 VLLM_KERNEL_SYMBOL_MAP=1`, two test steps).

- The image-build job produced `kernel_symbol_map.json.gz` (3.4 MB) on a real CI builder via the merged vLLM path.
- Both GPU steps uploaded `kern.<pid>.txt` plus the `kernrec.json` sidecar.
- `kernrec-collect` downloaded 4 recording artifacts and the map, built a v2 table (2 rows, both `passed`/`complete`, 0 drops, 275 unique kernels), validated `table usable: yes; symbol map usable: yes`, and uploaded both as artifacts of the collect job.
- Then, on the premerge agent, `aws s3 cp` returned AccessDenied and the step soft-failed with "S3 upload failed (bucket or write permission not in place?); artifacts are on this job". Nothing was written to the bucket. That is the intended behaviour off main: premerge stacks hold only ECR read.

The table artifact also loads through the selector-side reader in #623 (`load_table`: 2 rows, commit `91d7324c`, build 90406), so producer and consumer agree on the shape.

Possible follow-up: skip the S3 stage outright when `BUILDKITE_BRANCH != main`, so fork test builds do not show a red soft-fail. Small, separate.

